PR #32, opened 2020-01-10
Add Travis jobs for "examples" functional tests

DanielKamkha:
Updated build scripts for "examples" functional tests to fix a bug that
resulted Gluecodium itself being always built by those scripts even if
the corresponding command line option was not specified.

Updated .travis.yml file to create 4 jobs for functional tests (one for
C++, Android, Swift, and Dart each).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,4 @@
-language: java
-jdk:
-  - openjdk8
-sudo: false
+dist: bionic
 branches:
   except:
     - gh-pages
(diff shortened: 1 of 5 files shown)


examples/scripts/build-android
```diff
@@ -79,7 +79,7 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
-if [ $BUILD_LOCAL_GLUECODIUM ]; then
+if $BUILD_LOCAL_GLUECODIUM; then
```
teekahko: This will execute whatever is in the variable and branch based on the exit status. I think it would be safer to compare the variable contents instead.
`if [ "$BUILD_LOCAL_GLUECODIUM" = "true" ]; then ...`

DanielKamkha: Done

.travis.yml
```diff
@@ -16,8 +13,119 @@ before_cache:
...
+      install:
+        - CMAKE_VERSION=3.15.5
+        - wget -nv https://cmake.org/files/v${CMAKE_VERSION%.*}/cmake-${CMAKE_VERSION}-Linux-x86_64.tar.gz
+        - tar xf cmake-${CMAKE_VERSION}-Linux-x86_64.tar.gz > /dev/null
+        - mv cmake-${CMAKE_VERSION}-Linux-x86_64 cmake-install
+        - SWIFT_BRANCH=swift-5.1.3-release
+        - SWIFT_VERSION=swift-5.1.3-RELEASE
+        - SWIFT_PLATFORM=ubuntu16.04
+        - SWIFT_ARCHIVE_NAME=${SWIFT_VERSION}-${SWIFT_PLATFORM}
+        - SWIFT_URL=https://swift.org/builds/${SWIFT_BRANCH}/$(echo "${SWIFT_PLATFORM}" | tr -d .)/${SWIFT_VERSION}/${SWIFT_ARCHIVE_NAME}.tar.gz
```
Dschoordsch: Would be nice to have the variables first before other commands.

DanielKamkha: Done

Issue #83, opened 2020-01-16
Make JNI_OnLoad and JNI_OnUnload configurable


Dschoordsch:
As an Android developer I want to be able to inject custom hooks into `JNI_OnLoad` and `JNI_OnUnload`. Since these are global functions currently I cannot execute any code on `System.load_library`.

JniBase should allow to set the names of these load functions via `#define` through the build system so a library designer can have her own `JNI_OnLoad` function. It's then the library designers responsibility to call the renamed function from JniBase.

Dschoordsch:
JNI_OnUnload is not used by Gluecodium.

PR #88, opened 2020-01-16
Add possibility to redefine JNI_OnLoad

Dschoordsch:
Modified HelloWorld example to use custom JNI_OnLoad method.

Resolves: #83


gluecodium/src/main/resources/templates/jni/utils/JniBaseImplementation.mustache
```diff
@@ -25,6 +25,12 @@
 
 #include <pthread.h>
 
+// Allow to change JNI_OnLoad function name via build system since it is global and the user
+// might want to have custom implementation
+#ifndef GLUECODIUM_JNI_ONLOAD
+# define GLUECODIUM_JNI_ONLOAD
```
DanielKamkha: Why is this needed to be duplicated in the implementation file? Isn't it enough to have it in the header? If the header define is missing, this one will be empty and will produce uncompilable code regardless.

PR #99, opened 2020-01-21
Fix compilation issues in Optional template

rbuch703:
- the defined symbols for MSVC with C++17 could not be compiled as the
  code was not valid.
- the std::optional detection for non-MSVC compilers was fragile: The
  presence of an <optional> header does not guarantee that a compliant
  std::optional type exists. A compliant one can only be assumed to
  exist in C++17 compilers.

Signed-off-by: Robert Buchholz <robert.buchholz@here.com>

### DIFF
--- a/gluecodium/src/main/resources/templates/cpp/Optional.mustache
+++ b/gluecodium/src/main/resources/templates/cpp/Optional.mustache
@@ -43,11 +43,11 @@
 
 /*
  * This header is used to define the optional type used by Gluecodium. If nothing else is defined
- * it will use std::optional if available and fallback to a builtin optional implementation
+ * it will use std::optional if available and fallback to a built-in optional implementation
  * otherwise. To use custom optional implementation, GLUECODIUM_OPTIONAL_HEADER and GLUECODIUM_OPTIONAL_TYPE
  * need to be defined. The optional will be used in the internal namespace.
  *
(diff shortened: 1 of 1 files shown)


gluecodium/src/main/resources/templates/cpp/Optional.mustache
```diff
@@ -65,25 +65,16 @@
...
-                                          auto in_place = std::in_place; \
-                                          auto bad_access = std::bad_optional_access;
+#  if (defined(_MSC_VER) && _MSC_VER >= 1911 && _MSVC_LANG >= 201703L) || __cplusplus >= 201703L
+#    define GLUECODIUM_OPTIONAL_HEADER <optional>
+#    define GLUECODIUM_OPTIONAL_TYPE std::optional
+#    define GLUECODIUM_OPTIONAL_HELPER_TYPES using std::nullopt; \
+                                             using std::in_place; \
+                                             using std::bad_optional_access;
+#  else
+#    define GLUECODIUM_OPTIONAL_HEADER "OptionalImpl.h"
```
Dschoordsch: Why do we have this here if we have another check in l81

rbuch703: Good point, done (removed the version in line 81).

Issue #85, opened 2020-01-16
Add support for exceptions in Dart generator


DanielKamkha:

DanielKamkha:
Needs #103 implemented first, due to the structuring of functional tests.

DanielKamkha:
Also needs #92 implemented, because there's a regression test for combining immutable structs and exceptions.

PR #110, opened 2020-01-27
Add support for exceptions in Dart generator

DanielKamkha:
Updated Dart and FFI templates to support propagating errors as Dart
exceptions.

Moved "Blob" use case from "Errors" functional test to "Blob" functional
test.

Resolves: #85


gluecodium/src/main/resources/templates/dart/DartInterface.mustache
```diff
@@ -52,7 +52,11 @@ final __release_handle = __lib.nativeLibrary.lookupFunction<
     void Function(Pointer<Void>)
   >('{{resolveName "Ffi"}}_release_handle');
 
-class {{resolveName}}__Impl implements {{resolveName}}{
+{{#functions}}
```
Dschoordsch: filtering here instead of in the linked template would make this bit easier to understand

Issue #65, opened 2020-01-14
Add support for methods and constants on structs in Dart generator


DanielKamkha:
This includes support for custom struct constructors.

DanielKamkha:
Needs exceptions support #85 implemented first, due to how functional tests are structured.

PR #111, opened 2020-01-28
Add support for methods in structs to Dart generator

DanielKamkha:
Moved generated struct conversion function from a separate Dart file
into the same where Dart struct definition resides. This is required
since the all-fields constructor for a struct is "private" if custom
constructors are defined. As there is no "internal" visibility in Dart
language, the only way to access the "private" (actually file-private)
constructor is to put everything into the same file.

Updated Dart templates to support methods (and constructors) on structs.

Added/updated smoke and functional tests for both methods and constants
on structs.

Resolves: #65


examples/platforms/dart/test/StructsWithMethods_test.dart
```diff
@@ -0,0 +1,167 @@
...
+
+    try {
+      result = new Vector.createCopy(vector);
+    } on ValidationException catch(e) {
+      exception = e;
+    }
+
+    expect(result.x, 1);
+    expect(result.y, 2);
+    expect(exception, isNull);
```
Dschoordsch: -1 this should be before `expect(result.x, 1);` because result == null in case of exception

examples/platforms/dart/test/StructsWithMethods_test.dart
```diff
@@ -0,0 +1,167 @@
...
+    try {
+      result = new StructsWithMethodsInterface_Vector3.createCopy(vector3);
+    } on ValidationException catch(e) {
+      exception = e;
+    }
+
+    expect(result.x, 1);
+    expect(result.y, 2);
+    expect(result.z, 3);
+    expect(exception, isNull);
```
Dschoordsch: -1 Needs to be checked before accessing result.

PR #112, opened 2020-01-28
Add support for Date type in Dart generator

DanielKamkha:
Resolves: #62

### DIFF
--- a/gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartGeneratorSuite.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartGeneratorSuite.kt
@@ -207,18 +207,15 @@ class DartGeneratorSuite(options: Gluecodium.Options) : GeneratorSuite() {
                 "$LIB_DIR/$SRC_DIR_SUFFIX/_library_init.dart"
             ),
             GeneratedFile(
-                TemplateEngine.render("dart/DartBooleanConversion", templateData),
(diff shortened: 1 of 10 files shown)


gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartGeneratorSuite.kt
```diff
@@ -206,19 +206,16 @@ class DartGeneratorSuite(options: Gluecodium.Options) : GeneratorSuite() {
...
-                TemplateEngine.render("dart/DartStringConversion", templateData),
-                "$LIB_DIR/$SRC_DIR_SUFFIX/String__conversion.dart"
-            ),
             GeneratedFile(
                 TemplateEngine.render("dart/DartPubspec", templateData),
                 "$ROOT_DIR/pubspec.yaml"
             )
-        )
+        ) + listOf("Boolean", "String", "DateTime").map {
+            GeneratedFile(
```
Dschoordsch: Not important right now, but maybe you could already put this in the common source set? I don't know how dart build system handles it, but that should be for the build system to decide then.

DanielKamkha: Dart modularization is super weird, so I'd prefer to not future-proof this right now. There's a high chance a "solution" we pick right now might turn out wrong in the long run.

PR #114, opened 2020-01-28
Add support for Blob type to Dart generator

DanielKamkha:
Added Dart and FFI templates for Blob type conversion. Fixed types and
indentation in Dart template for Generic types conversion.

Resolves: #57

### DIFF
--- a/examples/platforms/dart/test/Blobs_test.dart
+++ b/examples/platforms/dart/test/Blobs_test.dart
@@ -18,6 +18,7 @@
 //
 // -------------------------------------------------------------------------------------------------
 
+import "dart:typed_data";
 import "package:test/test.dart";
 import "package:hello/hello.dart";
 import "../test_suite.dart";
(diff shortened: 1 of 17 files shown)


examples/platforms/dart/test/Blobs_test.dart
```diff
@@ -0,0 +1,79 @@
...
+    ExplosiveException exception = null;
+
+    try {
+      result = ArraysByteBuffer.methodThatExplodes(false);
+    } on ExplosiveException catch(e) {
+      exception = e;
+    }
+
+    expect(result, [0, 1, 2]);
+    expect(exception, isNull);
```
Dschoordsch: -1 please check for null exception first

gluecodium/src/main/resources/templates/ffi/FfiBlobHandleHeader.mustache
```diff
@@ -0,0 +1,44 @@
...
+#include "OpaqueHandle.h"
+#include <cstdint>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+_GLUECODIUM_FFI_EXPORT FfiOpaqueHandle blob_create_handle();
+_GLUECODIUM_FFI_EXPORT void blob_release_handle(FfiOpaqueHandle handle);
+_GLUECODIUM_FFI_EXPORT void blob_insert(FfiOpaqueHandle handle, uint8_t value);
```
Dschoordsch: There is no concept of a byte array in Dart where you could pass a raw pointer and length instead?

DanielKamkha: There's a ByteBuffer class: https://api.dart.dev/stable/2.7.1/dart-typed_data/ByteBuffer-class.html, but it does not seem to be intended for non-system use, as there's no sane way to create it.

There's also Uint8List, fixed-size List implementation optimized for 8-bit ints: https://api.dart.dev/stable/2.7.1/dart-typed_data/Uint8List-class.html. I think I should use it here, thanks for the hint.

gluecodium/src/main/resources/templates/ffi/FfiBlobHandleHeader.mustache
```diff
@@ -30,14 +30,11 @@
...
+_GLUECODIUM_FFI_EXPORT FfiOpaqueHandle blob_create_handle(uint64_t length);
 _GLUECODIUM_FFI_EXPORT void blob_release_handle(FfiOpaqueHandle handle);
-_GLUECODIUM_FFI_EXPORT void blob_insert(FfiOpaqueHandle handle, uint8_t value);
-_GLUECODIUM_FFI_EXPORT FfiOpaqueHandle blob_iterator(FfiOpaqueHandle handle);
-_GLUECODIUM_FFI_EXPORT void blob_iterator_release_handle(FfiOpaqueHandle iterator_handle);
-_GLUECODIUM_FFI_EXPORT bool blob_iterator_is_valid(FfiOpaqueHandle handle, FfiOpaqueHandle iterator_handle);
-_GLUECODIUM_FFI_EXPORT void blob_iterator_increment(FfiOpaqueHandle iterator_handle);
-_GLUECODIUM_FFI_EXPORT uint8_t blob_iterator_get(FfiOpaqueHandle iterator_handle);
+_GLUECODIUM_FFI_EXPORT uint64_t blob_get_length(FfiOpaqueHandle iterator_handle);
+_GLUECODIUM_FFI_EXPORT uint8_t blob_get_at(FfiOpaqueHandle iterator_handle, uint64_t index);
```
Dschoordsch: Already better, but I was wondering whether or not you could pass a pointer to the underlying ByteBuffer? It seems there is no easy way to assign Int8pointer content to a ByteBuffer in Dart, but you could get the list length from FFI, create UInt8List in Dart and then pass the underlying ByteBuffer to FFI to copy the values into it via std::copy or memcpy.

DanielKamkha: I'm not sure if this is even possible. I created a spike #115 to investigate this later.

Issue #93, opened 2020-01-20
Add support for struct field initializers in Dart generator


DanielKamkha:

DanielKamkha:
Needs #60 implemented first, due to functional tests entanglement.

PR #117, opened 2020-01-29
Add support for struct field initializers to Dart generator

DanielKamkha:
Removed "nullable" use case from "Defaults" functional test as it was a
duplicate of what is already tested in "Nullable" functional test.

Added smoke and functional tests as appropriate.

Resolves: #93

### DIFF
--- a/gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartNameResolver.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartNameResolver.kt
@@ -100,7 +100,10 @@ internal class DartNameResolver(
                 val postfix: String
                 when (actualType) {
                     is LimeStruct -> {
-                        prefix = "${resolveName(actualType)}("
+                        val useDefaultsConstructor =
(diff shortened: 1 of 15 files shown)


gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartNameResolver.kt
```diff
@@ -100,7 +100,10 @@ internal class DartNameResolver(
                 val postfix: String
                 when (actualType) {
                     is LimeStruct -> {
-                        prefix = "${resolveName(actualType)}("
+                        val useDefaultsConstructor =
+                            actualType.fields.isNotEmpty() && limeValue.values.isEmpty()
+                        val constructorName = if (useDefaultsConstructor) ".withDefaults" else ""
```
Dschoordsch: I think the name `.withDefaults` belongs in the name rules and ideally should be configurable.

DanielKamkha: Created #120 for this.

PR #119, opened 2020-01-30
Updated Swift building support

Dschoordsch:
TargetSources can now be used for Swift without adding the Swift files
directly to the target. This makes it possible to build one framework
consisting of different CMake targets by only compiling Swift code for
the final target. Generation itself and C++/C compilation is still
done individually.

Signed-off-by: Georg Bremer <georg.bremer@here.com>


cmake/modules/gluecodium/gluecodium/Generate.cmake
```diff
@@ -114,6 +114,7 @@ function(apigen_generate)
 
   message(STATUS "${operationVerb} '${APIGEN_TARGET}' with '${APIGEN_GENERATOR}' generator using Gluecodium version '${APIGEN_VERSION}'
   Input: '${apigen_generate_LIME_SOURCES}'")
+  message("GEORG target: ${APIGEN_TARGET} Input: '${apigen_generate_LIME_SOURCES}'")
```
DanielKamkha: Please remove

DanielKamkha:
Looks fine to me now, but I'd prefer if @Hsilgos would take a look as well.

cmake/modules/gluecodium/gluecodium/TargetSources.cmake
```diff
@@ -71,8 +76,60 @@ function(apigen_target_sources target)
...
+          PROPERTY GENERATED 1)
+
+      foreach(generated_file
+          ${BUILD_OUTPUT_DIR}/${APIGEN_GENERATED_cbridge_${_source_set}}
+          ${BUILD_OUTPUT_DIR}/${APIGEN_GENERATED_cbridge_header_${_source_set}}
+          ${BUILD_OUTPUT_DIR}/${APIGEN_GENERATED_swift_${_source_set}})
+        if(NOT EXISTS "${generated_file}")
+          get_filename_component(directory "${generated_file}" DIRECTORY)
+          file(MAKE_DIRECTORY "${directory}")
+          file(WRITE "${generated_file}" "// Dummy file to be replaced by Gluecodium during build, see also https://gitlab.kitware.com/cmake/cmake/issues/18399")
```
Hsilgos: If it's dummy file which will be replaced, may then it should contain #error "Dummy file blah blah ..."?

examples/scripts/build-ios-on-mac
```diff
@@ -61,7 +61,7 @@ function build {
         pushd "${build_dir}"
         safe cmake \
             -DBUILD_SHARED_LIBS=ON \
-            -DCMAKE_TOOLCHAIN_FILE="${SCRIPT_DIR}/../cmake/external/ios-cmake/ios.toolchain.cmake" \
+            -DCMAKE_TOOLCHAIN_FILE="${SCRIPT_DIR}/../../cmake/external/ios-cmake/ios.toolchain.cmake" \
```
Hsilgos: How about to try to use CMake's toolchain? Separate commit?

Dschoordsch: There used to be a ticket for this in another related repository.

cmake/modules/gluecodium/gluecodium/TargetSources.cmake
Dschoordsch: Yes, good idea.

Issue #123, opened 2020-01-31
Add basic support for calling Dart functions from C++ in Dart generator


DanielKamkha:
In-scope for this issue: calling Dart functions from C++ in a single-threaded scenario, with or without return values.

Out of scope: exception handling, multi-threading.

DanielKamkha:
Useful info:
https://github.com/dart-lang/sdk/issues/35761
https://github.com/dart-lang/sdk/issues/37301
https://github.com/dart-lang/sdk/issues/37022

PR #141, opened 2020-02-17
Add support for basic use cases of calling Dart code from C++

DanielKamkha:
Updated Dart and FFI templates to generate "proxy" C++ classes
representing callable Dart interfaces in C++. This supports basic use
cases of calling Dart callbacks from C++ in the same thread, with or
without return value. Error handling and multithreading will be done
separately.

Resolves: #123


gluecodium/src/test/resources/smoke/listeners/output/dart/ffi/ffi_smoke_CalculatorListener.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include <vector>
+#include <memory>
+#include <new>
+class smoke_CalculatorListener_Proxy : public ::smoke::CalculatorListener {
+public:
+    smoke_CalculatorListener_Proxy(uint64_t token, FfiOpaqueHandle f0, FfiOpaqueHandle f1, FfiOpaqueHandle f2, FfiOpaqueHandle f3, FfiOpaqueHandle f4, FfiOpaqueHandle f5)
+        : token(token), f0(f0), f1(f1), f2(f2), f3(f3), f4(f4), f5(f5) { }
+    void
+    on_calculation_result(const double calculationResult) override {
+        int64_t _error = (*reinterpret_cast<int64_t (*)(uint64_t, double)>(f0))(token,
```
Hsilgos: I would suppress warning about unused variable

DanielKamkha: That's not particularly important as it will get used in the next implementation.

PR #148, opened 2020-02-20
Adds CFBundleSupportedPlatforms with detected platform to plist for iOS

Hsilgos:
Sometimes Xcode can't export archive because it's not sure about target
platform. The option CFBundleSupportedPlatforms should list suported
platforms to resolve this issue.

This patch adds the option and simple logic to detect platform based
on enabled architecture (see CMake option CMAKE_OSX_ARCHITECTURES)

### DIFF
--- a/cmake/modules/gluecodium/swift/FrameworkInfoPlist.cmake
+++ b/cmake/modules/gluecodium/swift/FrameworkInfoPlist.cmake
@@ -23,6 +23,30 @@ set(includeguard_gluecodium_swift_FrameworkInfoPlist ON)
 cmake_minimum_required(VERSION 3.5)
 
 #.rst:
+# apigen_check_only_listed_ios_architectures_are_enabled
+# ---------------------------------
+#
+# Checks if only listed architectures are enabled with CMAKE_OSX_ARCHITECTURES
+#   variable
+#
+# .. command:: apigen_check_only_listed_ios_architectures_are_enabled
+#
+# The general form of the command is::
+#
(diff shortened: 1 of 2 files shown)


cmake/modules/gluecodium/swift/FrameworkInfoPlist.cmake
```diff
@@ -48,6 +72,20 @@ function(apigen_swift_framework_info_plist target)
   get_target_property(SWIFT_RESOURCES_DIR ${target} APIGEN_SWIFT_RESOURCES_DIR)
   get_target_property(SWIFT_FRAMEWORK_NAME ${target} APIGEN_SWIFT_FRAMEWORK_NAME)
   get_target_property(MACOSX_FRAMEWORK_IDENTIFIER ${target} APIGEN_SWIFT_FRAMEWORK_IDENTIFIER)
+  if(IOS)
+    if(CMAKE_OSX_ARCHITECTURES) # Support CMake's default toolchain
+      apigen_check_only_listed_ios_architectures_are_enabled (_is_iphone_os "armv7" "armv7s" "arm64" "arm64e")
```
DanielKamkha: Will this work for simulator? It has "x64" architecture, right?

Hsilgos: This function detects if all architectures listen in CMAKE_OSX_ARCHITECTURES are architectures for device. If not - below next branch will set APPLE_PLATFORM_NAME to simulator. So yes, this logic works for simulators

Issue #131, opened 2020-02-11
Preserve type information on Dart<-C++ transitions


DanielKamkha:
For the scenario of class or interface inheritance. When Child inherits from Parent, and a function returns Parent type, but the real instance return is Child, the resulting object should be castable to Child type. This should work both in Dart->C++ and C++->Dart directions.

PR #155, opened 2020-02-25
Preserve type information to Cpp-to-Dart transition

DanielKamkha:
Updated Dart and FFI templates to preserve type information to
Cpp-to-Dart transition for classes/interfaces that inherit from other
classes/interfaces.

Resolves: #131


gluecodium/src/test/resources/smoke/inheritance/output/dart/lib/src/_type_repository.dart
```diff
@@ -0,0 +1,20 @@
...
+import 'package:library/src/smoke/ChildClassFromClass.dart';
+import 'package:library/src/smoke/ChildClassFromInterface.dart';
+import 'package:library/src/smoke/ChildInterface.dart';
+import 'package:library/src/smoke/InternalChild.dart';
+import 'package:library/src/smoke/InternalParent.dart';
+import 'package:library/src/smoke/ParentClass.dart';
+import 'package:library/src/smoke/ParentInterface.dart';
+import 'dart:ffi';
+import 'package:ffi/ffi.dart';
+final Map<String, Function> typeRepository = {
```
Hsilgos: HashMap?

DanielKamkha: Map<> here is just a hint for the type resolution to help it disambiguate {} to a map (not a set) when/if it's empty. Real implementation is created by {} on the right side of assignment. According to Dart sources, it's LinkedHashMap by default.

Issue #167, opened 2020-02-26
Optionally prefix Swift conversion functions for collections with an additional prefix


DanielKamkha:
Consider modularized build use case. There's a base module (say, "core") defining a common type (say, "GeoCoordinates"). And then there are two modules "A" and "B" depending on "core" each using "List<GeoCoordinates>" somewhere. When all these modules are combined, there's a duplication of conversion functions for this List and the Swift code will fail to compile.

A straightforward solution of generating the collection conversion functions for the type ("GeoCoordinates") immediately when the type is declared (as opposed to the current approach of doing so only when that collection is used) might work for List<> and Set<>. But it would not work for Map<>, because there are too many possible key-value-type combinations that may include "GeoCoordinates".

An alternative approach would be to optionally prefix the names of the conversion functions (both in Swift and CBridge layers) with an additional prefix, passed as a Gluecodium option in the command line. This will result is some minor code duplication, but will not fail Swift compilation.

PR #172, opened 2020-02-27
Add optional prefix to collection conversion functions in Swift

DanielKamkha:
Updated Swift templates to add an optional prefix to the names of
collection conversion functions. This is needed for modularization
support (#167). CBridge part will be done separately.

Added a command line parameter for this prefix and CMake support for
this command line parameter.

Added "libhello_" Swift prefix to functional tests to test this
functionality.


cmake/modules/gluecodium/gluecodium/Generate.cmake
```diff
@@ -88,6 +88,7 @@ function(apigen_generate)
       CPP_EXPORT
       CPP_NAMERULES
       SWIFT_NAMERULES
+      SWIFT_INTERNAL_PREFIX
```
Hsilgos: How about to detect prefix from target's name if not set?

DanielKamkha: I don't think it's a good idea, there should still be an option to go with no prefix. The prefix is only needed for modularized build, maybe not even always for those. Definitely not needed for monolith builds, but makes call stacks harder to read.

Issue #160, opened 2020-02-26
C++: Use move semantic in Gluecodium structs


haraldF:
When generating C++ code, e.g. for the following sample lime file:

```
struct Pet {
    PhotoUrls: List<String>
}
```
The output constructor looks like:

```C++
Pet::Pet( const ::std::vector< ::std::string >& photo_urls )
    : photo_urls( photo_urls )
{
}
```

This is very sub-optimal - the constructor takes parameters as const references, although it takes ownership of the parameters. This means that all the input parameters will be deep-copied.

The C++11 way to do this is to take all parameters by value, not by const reference, so the caller can choose whether to copy his input parameters or move them.

My suggestion for the generated code:

```C++
Pet::Pet( ::std::vector< ::std::string > photo_urls )
    : photo_urls( std::move(photo_urls) )
{
}
```

This will prevent a lot of needless copies of values and should result in more optimized C++ code.

PR #188, opened 2020-03-04
Use move semantic in C++ struct constructors

DanielKamkha:
Previous implementation for C++ struct constructors took parameters as
const references, even though it took ownership of the parameters. This
led to deep-copying of the values.

Updated C++ templates to employ move semantics instead: to take all
parameters by value, and them using std::move on the values.

Resolves: #160


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Gluecodium project Release Notes
 
+## Unreleased
+### Bug fixes:
```
Dschoordsch: I would consider this an improvement, a.k.a feature

DanielKamkha: Done

Issue #37, opened 2020-01-14
Improve generated documentation for @FunctionalInterface interfaces


DanielKamkha:
[Original issue by @datasun ]

```kotlin
// Callback passed to calculateRoute() in RoutingEngine.
// This function is called after completion of calculateRoute() in RoutingEngine.
// It is always called on the main thread.
// The first argument is the error in case of failure. It is `null` for an operation that succeeds.
// The second argument is the calculated routes. It is `null` in case of an error.
lambda CalculateRouteCallback = (RoutingError?, List<Route>?) -> Void
```
results to:
```kotlin
@FunctionalInterface
public interface CalculateRouteCallback {
    void apply(RoutingError var1, List<Route> var2);
}
```
1) Having a method called "apply" is not acceptable for Java.
2) Furthermore, names for parameters are broken: apply(RoutingError p0, java.util.List<Route> p1)
3) Resulting documentation looks ugly => The parameters are documented in one block. I would expect that each parameter gets documented like it's common for Javadoc.
4) Nullable expectations are missing in API Reference and in decompiled code.

For Java the documentation is expected to look like for any other interface or method. In Java there is no difference.

DanielKamkha:
Method renaming can be done with @Java(FunctionName = "foo") attribute in IDL file. Nullability issue is already fixed separately. The only remaining issue here generating proper JavaDoc comment for parameter documentation.

PR #191, opened 2020-03-05
Improve generated documentation for @FunctionalInterface in Java

DanielKamkha:
Updated LimeLambda and LimeLambdaParameter classes to allow for
representing dedicated comments on lambda's parameters and return type.

Updated JavaModelBuilder to propagate these dedicated comments into a
structured JavaDoc comment on the method of the lambdas's
@FunctionalInterface.

Updated "Comments" smoke test to include a use case for structured
comments on a lambda.

Resolves: #37

### DIFF
--- a/gluecodium/src/main/java/com/here/gluecodium/generator/java/JavaModelBuilder.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/generator/java/JavaModelBuilder.kt
@@ -56,6 +56,7 @@ import com.here.gluecodium.model.lime.LimeField
 import com.here.gluecodium.model.lime.LimeFunction
 import com.here.gluecodium.model.lime.LimeInterface
 import com.here.gluecodium.model.lime.LimeLambda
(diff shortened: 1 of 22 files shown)


gluecodium/src/main/java/com/here/gluecodium/generator/java/JavaModelBuilder.kt
```diff
@@ -335,23 +336,27 @@ class JavaModelBuilder(
...
-            JavaParameter(
+            val result = JavaParameter(
                 nameRules.getName(parameter, index),
                 typeMapper.applyNullability(
                     typeMapper.mapType(parameter.typeRef),
                     parameter.typeRef.isNullable
                 )
             )
+            result.comment = createComments(parameter)
+            result
```
Hsilgos: Is it leftover or funny syntax?

DanielKamkha: It is "funny syntax" to return a value from a block.

PR #204, opened 2020-03-16
Add support for @Equatable interfaces in C++

DanielKamkha:
Added smoke and functional tests for @Equatable interfaces in C++. Since
the generated code for interfaces and classes is identical for C++
generator, no changes to Mustache templates is necessary to support
@Equatable interfaces in C++.


DanielKamkha:
> Changelog entry?

Sure, but is is at the end of the commit chain, when all generators are covered.

PR #208, opened 2020-03-17
Additional CMake support for Dart fused builds

DanielKamkha:

### DIFF
--- a/cmake/modules/gluecodium/gluecodium/GeneratedSources.cmake
+++ b/cmake/modules/gluecodium/gluecodium/GeneratedSources.cmake
@@ -41,9 +41,14 @@ function(apigen_list_generated_sources _sources_list_output)
   cmake_parse_arguments(apigen_list_generated_sources "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
   apigen_set_generated_files(${apigen_list_generated_sources_TARGET})
-  if(NOT apigen_list_generated_sources_GENERATOR IN_LIST APIGEN_SUPPORTED_GENERATORS)
-    message(FATAL_ERROR "Generator ${apigen_list_generated_sources_GENERATOR} is not in list of supported generators: ${APIGEN_SUPPORTED_GENERATORS}")
-  endif()
+
+  string(REPLACE "," ";" _generators ${apigen_list_generated_sources_GENERATOR})
+  foreach(_generator ${_generators})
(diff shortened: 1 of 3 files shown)


cmake/modules/gluecodium/gluecodium/GeneratedSources.cmake
```diff
@@ -41,9 +41,14 @@ function(apigen_list_generated_sources _sources_list_output)
   cmake_parse_arguments(apigen_list_generated_sources "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
   apigen_set_generated_files(${apigen_list_generated_sources_TARGET})
-  if(NOT apigen_list_generated_sources_GENERATOR IN_LIST APIGEN_SUPPORTED_GENERATORS)
-    message(FATAL_ERROR "Generator ${apigen_list_generated_sources_GENERATOR} is not in list of supported generators: ${APIGEN_SUPPORTED_GENERATORS}")
-  endif()
+
+  string(REPLACE "," ";" _generators ${apigen_list_generated_sources_GENERATOR})
+  foreach(_generator ${_generators})
+    if(NOT _generator IN_LIST APIGEN_SUPPORTED_GENERATORS)
```
Hsilgos: Actually it's necessary too declare
`cmake_minium_required (3.3)`
otherwise IN_LIST may work weird

cmake/modules/gluecodium/gluecodium/runGenerate.cmake
```diff
@@ -72,15 +72,17 @@ endfunction()
 function(_collect_all_files_in_single_compilation_units)
     _include_all(cpp "cpp/*.cpp")
 
-    if(APIGEN_GENERATOR STREQUAL android)
+    if(APIGEN_GENERATOR MATCHES android)
```
Hsilgos: Probably `android IN_LIST APIGEN_GENERATOR` is clearer

cmake/modules/gluecodium/gluecodium/GeneratedSources.cmake
DanielKamkha: Well we do require even higher version elsewhere, should be fine.

cmake/modules/gluecodium/gluecodium/runGenerate.cmake
DanielKamkha: But it's not a CMake list, it's comma-separated, not semicolon-separated.

Issue #48, opened 2020-01-14
Create example project


DanielKamkha:
[Original issue by @Dschoordsch ]

As a new user of Gluecodium setting up a new project, there is a lot of things that can go wrong. I would like to simply copy an existing project and follow the comments inside.

Currently hello world example doesn't serve this purpose:

hello world CMake.txt is not standalone but requires two additional lines from the CMake file in parent folder, that should be fixed.
requires apigen CMake modules which is not documented (also not how to obtain those)
requires external build scripts to set the generator (should be a documented CMake option instead) is missing comments on what to change

Existing Hello World project should be deleted (while preserving functional tests as a non-UI test suite).

linsang21:
Hi there, I'm trying to check the progress on creating the example project. The project itself is well documented, but it'd be much helpful for me to have some reference on how to integrate with the existing project.

To be specifically, what I'm looking for is that I have a C++ project built by CMake into a static library and public C++ headers, which I'd like to export as Swift version so can be used on MacOS/ iOS.

Hsilgos:
Hi @linsang21 , there are plenty of samples in cmake [tests](https://github.com/heremaps/gluecodium/tree/master/cmake/tests/unit) were you may choose the variant of project configuration which suits you. This is good entry point to start. Reading your description you need something very basic like [this](https://github.com/heremaps/gluecodium/tree/master/cmake/tests/unit/gluecodium_generate/single-module-framework), but may be framework should be static.
After initial setup you may refer [functional tests](https://github.com/heremaps/gluecodium/tree/master/functional-tests/functional) which contain a lot of lime file [examples](https://github.com/heremaps/gluecodium/tree/master/functional-tests/functional/input/lime).

linsang21:
Thanks @Hsilgos. I was able to figure out the setup after referencing tests. There's one error I want to point out when using cmake.

The [build gradle](https://github.com/heremaps/gluecodium/blame/fc781dc45d20dfb6cfc33bf471de8309a5e0385e/cmake/modules/gluecodium/gluecodium/details/build.gradle#L30) seems out of date for getting the library. I had to changed to below to make build work.

```
...
repositories {
    mavenLocal()
    mavenCentral()
    maven { url 'https://jitpack.io' } // Added for gluecodium
    google() // Added for  com.android.tools.build:gradle:4.1.2
}

apply plugin: 'application'

dependencies {
    runtime "com.github.heremaps:gluecodium:$version" // Changed path to github
}
...
```

[...]

PR #270, opened 2020-04-07
Remove iOS example app

DanielKamkha:
iOS example app was not maintained for at least 2 years. It does not
compile anymore. It has to be removed and then replaced with a simpler
example app (see #48).


DanielKamkha:
Postponed. Needs to be validated on a Mac.

Issue #231, opened 2020-03-27
Add validation for method overloads in Dart


DanielKamkha:
Since Dart language does not have the concept of method overloads, each method and constructor overload in the IDL files need to be explicitly assigned a Dart-specific @Dart() name by the IDL file author.

Since Dart compilation is quirky, building a consistent set of smoke tests to detect that in project's build system is challenging. Providing a forewarning at Gluecodium generation step would be therefore helpful.

Validation warnings should be issued for overloaded methods and constructors without @Dart() platform names, as well as for overloaded constructors with multiple @Dart(Default) attributes. In a single set of overloads, one non-renamed method/constructor is fine. The warnings should only be issued if Dart generator is enabled for the current Gluecodium run.

Warnings should be issued instead of errors because it's Dart compiler's responsibility to emit these as errors. Gluecodium's validation in this case is just a supplementary warning system, to slightly relax the burden of Dart build system on the user, not to replace it completely.

Issue #232, opened 2020-03-27
Add a command line parameter to treat validation warnings as errors


DanielKamkha:
It would be helpful to have a command line parameter to treat validation warnings as errors. However, this needs to be configurable per type of validation warning. For example, "broken" documentation links are quite normal and might just mean web links inserted into the documentation. Blindly treating these as errors would be useless, so a more fine-grained configuration is required.

PR #272, opened 2020-04-09
Add validation for Dart method overloads

DanielKamkha:
Added DartOverloadsValidator to issue validation warnings for overloaded methods and constructors without @Dart() platform names, as well as for overloaded constructors with multiple @Dart(Default) attributes. In a single set of overloads, one non-renamed method/constructor is fine.

Warnings are issued instead of errors because it's Dart compiler's responsibility to emit these as errors. There are many more such compiler-detectable issues (e.g. language-invalid identifier names). Detecting all of them on Gluecodium side would have added a huge maintenance burden with only a minor upside. Detecting just some of them but not all would make Gluecodium validation inconsistent.

For this particular case, Gluecodium's validation is a supplementary warning system, added on a user's request to partially ease the burden on Dart build system quirks. It is up to the user to treat these warnings as errors (see https://github.com/heremaps/gluecodium/issues/232).

Resolves: #231


gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartOverloadsValidator.kt
```diff
@@ -0,0 +1,77 @@
...
+    private fun validateContainer(limeContainer: LimeContainerWithInheritance): Boolean {
+        val allFunctions = limeContainer.functions + limeContainer.inheritedFunctions
+        val constructors = allFunctions.filter { it.isConstructor }
+
+        val overloadedFunctions = (allFunctions - constructors)
+            .groupBy { nameResolver.resolveName(it) }
+            .filter { it.value.size > 1 }
+        overloadedFunctions.forEach { entry ->
+            val pathsString = entry.value.map { it.path.toString() }.sorted().joinToString()
+            logger.warning(
```
Hsilgos, line 54: Isn't it error?

gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartOverloadsValidator.kt
```diff
@@ -0,0 +1,77 @@
...
+        val overloadedConstructors =
+            constructors.groupBy { getConstructorName(it) }.filter { it.value.size > 1 }
+        overloadedConstructors.forEach { entry ->
+            val pathsString = entry.value.map { it.path.toString() }.sorted().joinToString()
+            val warningText = if (entry.key.isEmpty()) {
+                "there is more than one default constructor: $pathsString"
+            } else {
+                "there is more than one constructor with '${entry.key}' name: $pathsString"
+            }
+            logger.warning(entry.value.first(), warningText)
```
Hsilgos, line 69: the same

gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartOverloadsValidator.kt
DanielKamkha: I updated the commit message with an explanation.

gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartOverloadsValidator.kt
DanielKamkha: I updated the commit message with an explanation.

Issue #268, opened 2020-04-07
Introduce @Cpp(CString) attribute for parameters in IDL


hanneswuerfel:
Currently a method overload like this:
```
    fun foo(
        bar: String
    ):
    fun foo(
        bar: Long
    ):
```
would result in C++ code produced like this:
```
virtual ::std::shared_ptr< Foo > foo( const ::std::string& bar  ) = 0;
virtual ::std::shared_ptr< Foo > foo( const int64_t bar ) = 0;
```
When calling:
`Foo::foo( "bar" ); // assignes value "1", not "bar"`
because the int64_t overload is used.
I want to create a const char* value overload through the lime file in order for users not to write:
`Foo::foo( std::string("bar") );`

PR #274, opened 2020-04-09
Add support for C strings in C++ generated code

DanielKamkha:
Added new LimeIDL attribute `@Cpp(CString)`. This attribute marks a String
type reference in IDL to be treated in C++ as `const char*` instead
of `std::string`. It has no effect on other types.

Updated C++, Swift, Java, and Dart generators and templates to support this.

Resolves: #268


gluecodium/src/main/resources/templates/cbridge/BaseHandleImpl.mustache
```diff
@@ -263,3 +264,28 @@ struct Conversion< std::chrono::system_clock::time_point >
...
+        return reinterpret_cast< std::string* >( ref )->c_str();
+    }
+
+    static const char*
+    toCppReturn( _baseRef ref )
+    {
+        auto ptr = reinterpret_cast< std::string* >( ref );
+        std::string value( *ptr );
+        delete ptr;
+        return value.c_str();
```
Hsilgos, line 289: It should be crash. 'value' runs out of scope and destroyed

gluecodium/src/main/resources/templates/cbridge/BaseHandleImpl.mustache
```diff
@@ -263,3 +264,28 @@ struct Conversion< std::chrono::system_clock::time_point >
...
+    static const char*
+    toCpp( _baseRef ref )
+    {
+        return reinterpret_cast< std::string* >( ref )->c_str();
+    }
+
+    static const char*
+    toCppReturn( _baseRef ref )
+    {
+        auto ptr = reinterpret_cast< std::string* >( ref );
```
Hsilgos, line 286: I would use unique_ptr:
std::unique_ptr ptr(reinterpret_cast< std::string* >( ref ));

gluecodium/src/main/resources/templates/jni/utils/JniCppConversionUtilsImplementation.mustache
```diff
@@ -42,6 +42,12 @@ convert_string_from_jni( JNIEnv* env, jstring java_string )
     return nresult;
 }
 
+const char*
+convert_cstring_from_jni( JNIEnv* env, jstring java_string )
+{
+    return env->GetStringUTFChars( java_string, nullptr );
```
Hsilgos, line 48: Smells like memory leak. Who will clean this up? ReleaseStringUTFChars must be called after use.

DanielKamkha:
> I don't think that it's good idea to provide functionality to pass "const char*" in both directions, it may lead to memory leaks. Initial issue was about overloading of methods to work both with std::string and "const char*" and not about receiving it back, so Gluecodium should just generate two methods: "std::string" and "const char*" as i understand the task.

Yeah, you're right. The bindings code for const char* looks awful. Just producing the overload for the purposes of C++ API only and never using it in bindings sounds much safer. I'll re-implement this.

Issue #227, opened 2020-03-26
Allow documentation links to method overloads


pbula:
There is currently no way to have links to overloaded methods in generated reference documentation comments. This is an especially useful feature when an overloaded method is introduced and the old one is deprecated. Putting a link to the new version in the deprecation message would be nice to have.

See attached Lime file for an example.
[Overloads.lime.txt](https://github.com/heremaps/gluecodium/files/4387976/Overloads.lime.txt)

Note how the deprecation message in `InterfaceWithOverloadedMethods` has a `{@link com.here.com.example.InterfaceWithOverloadedMethods#overloadedMethod}` javadoc link generated. This essentially links to itself. To have a proper deprecation message, the link should look like this: `{@link com.here.com.example.InterfaceWithOverloadedMethods#overloadedMethod(double)}`.

PR #275, opened 2020-04-09
Added support for including method signature in doc links.

DanielKamkha:
Updated AbstractLimeBasedModelBuilder class and its usages to support
referring to methods by name + signature (e.g. foo(String, Boolean)) in
documentation comment links. This enables distinguishing between
different method overloads of the same name (although the target
language generators cannot make this distinction yet, see #227).

### DIFF
--- a/gluecodium/src/main/java/com/here/gluecodium/generator/common/modelbuilder/AbstractLimeBasedModelBuilder.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/generator/common/modelbuilder/AbstractLimeBasedModelBuilder.kt
@@ -29,6 +29,7 @@ import com.here.gluecodium.model.lime.LimeContainerWithInheritance
 import com.here.gluecodium.model.lime.LimeElement
 import com.here.gluecodium.model.lime.LimeEnumerator
 import com.here.gluecodium.model.lime.LimeException
+import com.here.gluecodium.model.lime.LimeFunction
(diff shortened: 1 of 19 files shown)


gluecodium/src/main/java/com/here/gluecodium/generator/common/modelbuilder/AbstractLimeBasedModelBuilder.kt
```diff
@@ -55,13 +57,19 @@ abstract class AbstractLimeBasedModelBuilder<E>(
...
+     * disambiguation suffix is omitted) and full-signature function references are also allowed.
      */
     protected fun storeNamedResult(limeElement: LimeNamedElement, element: E) {
         storeResult(element)
         referenceMap[limeElement.fullName] = element
         val ambiguousKey = limeElement.path.withSuffix("").toString()
         referenceMap[ambiguousKey] = element
+        if (limeElement is LimeFunction) {
+            val signatureKey = ambiguousKey + limeElement.parameters
+                .joinToString(prefix = "(", postfix = ")") { it.typeRef.toString() }
```
Dschoordsch, line 69: I think it would be better using `separator = ","` and stripping whitespace for lookup to avoid being sensitive to sane formatting of multiple parameters.

DanielKamkha: Done

PR #280, opened 2020-04-15
Add logo to the readme

Dschoordsch:
Put the logo after the first introduction, first introduction is the most important part to read while the logo also gives a clue what this could be about.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ eliminating the need for writing conversions and bindings manually.
 To simplify use of platform specific functions, Gluecodium is also able to generate interfaces/protocols
 allowing use of Java/Swift/Dart implementations from C++.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -7,6 +7,8 @@ eliminating the need for writing conversions and bindings manually.
 To simplify use of platform specific functions, Gluecodium is also able to generate interfaces/protocols
 allowing use of Java/Swift/Dart implementations from C++.
 
+![Gluecodium logo](docs/logo.svg)
```
DanielKamkha, line 10: This is just too huge. Does not even fit onto my laptop screen vertically (since there are also UI elements from GitHub and the browser). Can it be downsized in Markdown? If not, we would need a separate file with a smaller logo.

Issue #46, opened 2020-01-14
Preserve identity of interface instances on platform side


DanielKamkha:
[Original issue by Piotr Bula]

As a developer, I want to be able use instances of Gluecodium generated interfaces the same way I would use regular java/swift objects. In particular, I want to be able to test the objects for their identity. I want to be able use them as keys in a map, put them in sets and so on.

This is currently not possible, as each time an object is returned from C++ side, it's a completely new java/swift object and as such it is impossible to test if two instances actually point to the same C++ object.

This forced us to add artificial unique ID to some of our classes, since without the ability to identify instances, their usefulness is very limited.

DanielKamkha:
This would most probably make @PointerEquatable attribute obsolete, as all classes and interfaces will become pointer equatable in both directions.

PR #299, opened 2020-04-23
Preserve referential equality in Java code

DanielKamkha:
Added JavaWrapperCache template to cache Java objects created to wrap
instances passed from C++ to Java. Updated other Java templates to use
this cache to preserve referential equality (i.e. the same C++ instance
passed to Java twice will be represented by the same Java object, see #46).

### DIFF
--- a/gluecodium/src/test/resources/smoke/generic_types/output/android/jni/com_example_smoke_GenericTypesWithGenericTypes.cpp
+++ b/gluecodium/src/test/resources/smoke/generic_types/output/android/jni/com_example_smoke_GenericTypesWithGenericTypes.cpp
@@ -6,6 +6,7 @@
 #include "ArrayConversionUtils.h"
 #include "JniClassCache.h"
 #include "JniReference.h"
+#include "JniWrapperCache.h"
 extern "C" {
 jobject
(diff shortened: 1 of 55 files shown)


gluecodium/src/test/resources/smoke/generic_types/output/android/jni/com_example_smoke_GenericTypesWithGenericTypes.cpp
```diff
@@ -115,6 +116,8 @@ Java_com_example_smoke_GenericTypesWithGenericTypes_methodWithMapGenericKeys(JNI
 JNIEXPORT void JNICALL
 Java_com_example_smoke_GenericTypesWithGenericTypes_disposeNativeHandle(JNIEnv* _jenv, jobject _jinstance, jlong _jpointerRef)
 {
-    delete reinterpret_cast<std::shared_ptr<::smoke::GenericTypesWithGenericTypes>*> (_jpointerRef);
+    auto p_nobj = reinterpret_cast<std::shared_ptr<::smoke::GenericTypesWithGenericTypes>*>(_jpointerRef);
```
Hsilgos, line 119: I would protect such code with unique_ptr

DanielKamkha: Seems unnecessary as this is the only code path that could delete the shared_ptr.

PR #300, opened 2020-04-24
Fixes xompilation issue with Xcode bevause of broken compilation order.

Hsilgos:
find_program was looking in wrong place for 'touch' utility
when cross compilation is enabled. Solution - don't use it and just
call utility directly and in case of some issue - treat it as
WARNING


limingchina:
Typo for commit message? “Compilation” “because”

PR #305, opened 2020-04-27
Preserve referential equality in Swift code

DanielKamkha:
Added WrapperCache.h static (non-template) implementation to cache Swift
objects created to wrap instances passed from C++ to Swift. Updated
other Swift templates to use this cache to preserve referential equality
(i.e. the same C++ instance passed to Swift twice will be represented by
the same Swift object, see #46).

### DIFF
--- a/gluecodium/src/main/resources/cbridge_internal/include/WrapperCache.h
+++ b/gluecodium/src/main/resources/cbridge_internal/include/WrapperCache.h
@@ -1,0 +1,64 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (C) 2016-2020 HERE Europe B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+// License-Filename: LICENSE
+//
+// -------------------------------------------------------------------------------------------------
+
+#pragma once
+
+#include <atomic>
+#include <mutex>
+#include <unordered_map>
+
+static std::atomic<bool> wrapper_cache_is_alive;
+
+namespace
+{
+class WrapperCache {
+public:
+    using CppPtr = const void*;
+    using SwiftPtr = const void*;
+
+    explicit WrapperCache() noexcept { wrapper_cache_is_alive = true; }
+    ~WrapperCache() { wrapper_cache_is_alive = false; }
+
+    void
+    cache_wrapper(CppPtr cpp_ptr, SwiftPtr swift_ptr) {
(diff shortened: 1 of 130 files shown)


gluecodium/src/main/resources/cbridge_internal/include/WrapperCache.h
```diff
@@ -0,0 +1,72 @@
...
+#pragma once
+
+#include <mutex>
+#include <unordered_map>
+
+namespace
+{
+class WrapperCache {
+public:
+    bool alive;
```
Hsilgos, line 30: According to the logic should be static? And i'd do it atomic (to decrease chance to receive cached value).
Honestly such approach hardly helps.

gluecodium/src/main/resources/cbridge_internal/include/WrapperCache.h
```diff
@@ -0,0 +1,72 @@
...
+
+    using CppPtr = const void*;
+    using SwiftPtr = const void*;
+
+    WrapperCache() : alive(true) {}
```
Hsilgos, line 35: mark it noexcept

gluecodium/src/main/resources/cbridge_internal/include/WrapperCache.h
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// Handle cases of non-deterministic ordering of construction and destruction of global
+// variables. This is on two ends:
+// 1. Use function static variable to ensure it's constructed on first use.
+// 2. Use an "alive" flag to ensure it won't crash if destructed early. (e.g. storing a
+//    shared_ptr to a proxy object as a static variable)
+static WrapperCache& get_wrapper_cache()
+{
+    static WrapperCache cache;
```
Hsilgos, line 70: I'd move implementation of get_wrapper_cache to cpp if possible (only 'common' code must contain it)

gluecodium/src/main/resources/cbridge_internal/include/WrapperCache.h
DanielKamkha: Making it atomic is reasonable. Making it static is not necessary, since the class itself is not a "static class". The cache is a singleton, yes, but that's implemented through get_wrapper_cache() below.

DanielKamkha: As for the approach itself, I don't like it either. I copied it from CachedProxyBase.h, where it was introduced by HARP team, supposedly to work around some static destruction order mishap. No idea why this never was a problem for other projects, but looks relatively harmless, so let's keep it.

gluecodium/src/main/resources/cbridge_internal/include/WrapperCache.h
DanielKamkha: Done.

gluecodium/src/main/resources/cbridge_internal/include/WrapperCache.h
Hsilgos: > Making it static is not necessary, since the class itself is not a "static class". The cache is a singleton, yes, but that's implemented through get_wrapper_cache() below.

May then extract variable from class? It's kind of UB when you use memory allocated for static class after destroying of the object.

gluecodium/src/main/resources/cbridge_internal/include/WrapperCache.h
DanielKamkha: Done

PR #311, opened 2020-04-30
Synchronises invocations of Gradle

Hsilgos:
Parallel invocations lead to unexpected behaviour
There was locking file which is already used to
synchronise generation of code. The same locking
file is now used to synchronise compilation of Java code

### DIFF
--- a/cmake/modules/gluecodium/gluecodium/Gradle.cmake
+++ b/cmake/modules/gluecodium/gluecodium/Gradle.cmake
@@ -1,0 +1,39 @@
+# Copyright (C) 2016-2020 HERE Europe B.V.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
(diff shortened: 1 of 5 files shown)


cmake/modules/gluecodium/gluecodium/Gradle.cmake
```diff
@@ -0,0 +1,39 @@
+# Copyright (C) 2016-2019 HERE Europe B.V.
```
DanielKamkha, line 1: 2016-2020

cmake/modules/gluecodium/java/runCompile.cmake
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env cmake -P
+# Copyright (C) 2016-2019 HERE Europe B.V.
```
DanielKamkha, line 2: 2016-2020

cmake/modules/gluecodium/java/runCompile.cmake
```diff
@@ -0,0 +1,62 @@
...
+    -b=compileJava.gradle
+    -PsrcDirs=${_java_source_dirs}
+    -PoutputDir=${APIGEN_JAVA_COMPILE_OUTPUT_DIR}
+    -PlocalDependencies=${APIGEN_JAVA_LOCAL_DEPENDENCIES}
+    -PlocalDependenciesDirs=${APIGEN_JAVA_LOCAL_DEPENDENCIES_DIRS}
+    -PlocalJars=${APIGEN_JAVA_LOCAL_JARS}
+    -PremoteDependencies=${APIGEN_JAVA_REMOTE_DEPENDENCIES}
+    compileJava
+  WORKING_DIRECTORY ${APIGEN_GLUECODIUM_DIR}
+  RESULT_VARIABLE COMPILE_RESULT
```
DanielKamkha, line 52: Could we please not do this output redirection unconditionally? I understand that for CI usage a log file is more useful. But these scripts are also used for Gluecodium's own development and for this immediate console output is more usefult.

Please make this configurable, defaulting to console output it not configured.

PR #320, opened 2020-05-05
Reduces long command line on Windows due to limitation of 8196 symbols

Hsilgos:
CMake puts post build events to batch file in Windows but
batch files use cmd.exe which has limitation of 8196 symbols

### DIFF
--- a/cmake/modules/gluecodium/gluecodium/Generate.cmake
+++ b/cmake/modules/gluecodium/gluecodium/Generate.cmake
@@ -103,10 +103,10 @@ function(apigen_generate)
   _apigen_set_option_or_default(VERSION +)
 
   set(operationVerb "Generate")
-  set(validateParam "")
+  set(validateProperty "false")
   if(${apigen_generate_VALIDATE_ONLY})
     set(operationVerb "Validate")
-    set(validateParam "-validateOnly")
+    set(validateProperty "true")
   endif()
 
   message(STATUS "${operationVerb} '${APIGEN_TARGET}' with '${APIGEN_GENERATOR}' generator using Gluecodium version '${APIGEN_VERSION}'
@@ -123,12 +123,13 @@ function(apigen_generate)
     set(apigen_generate_GENERATOR "cpp,${apigen_generate_GENERATOR}")
   endif()
 
-  set(APIGEN_GLUECODIUM_ARGS "\
-    -output \"${APIGEN_OUTPUT_DIR}\"\
-    -generators ${apigen_generate_GENERATOR}\
-    ${validateParam}\
-    ${mergeManifest}\
-    --enable-caching")
+  set(APIGEN_GLUECODIUM_PROPERTIES "\
+output=${APIGEN_OUTPUT_DIR}\n\
+generators=${apigen_generate_GENERATOR}\n\
+validate=${validateProperty}\n\
+cache=true\n")
+
+  unset (_apigen_input_list)
   foreach(input ${apigen_generate_LIME_SOURCES})
     # Attach sources to target for IDEs to display them properly in their projects
     file(GLOB_RECURSE inputLimeSources ${input}/*.lime)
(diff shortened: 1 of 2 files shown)


cmake/modules/gluecodium/gluecodium/Generate.cmake
```diff
@@ -185,11 +185,14 @@ function(apigen_generate)
   set (_lime_interface_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},INTERFACE_SOURCES>,INCLUDE,.*\\.lime$>")
   set (_lime_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},SOURCES>,INCLUDE,.*\\.lime$>")
 
+  set (_gluecodium_args_file "${CMAKE_CURRENT_BINARY_DIR}/gluecodium-command-line.txt")
```
Dschoordsch, line 188: This path cannot conflict with other Genium runs?

Hsilgos: Makes sense in case when one folder contains few calls to apigen_generate.
I'll think about some magic there to make name more unique

cmake/modules/gluecodium/gluecodium/runGenerate.cmake
```diff
@@ -51,10 +75,12 @@ function(_generate)
...
     endif()
 
-    list(FILTER APIGEN_AUX_FILES INCLUDE REGEX ".*\\.lime$")
-    foreach(_aux_lime IN LISTS APIGEN_AUX_FILES)
-      string(APPEND APIGEN_GLUECODIUM_ARGS " -auxinput \"${_aux_lime}\"")
-    endforeach()
+    file(READ "${APIGEN_GLUECODIUM_OPTIONS_FILE}" _options_content)
+    message(STATUS "Gluecodium options (stored in file ${APIGEN_GLUECODIUM_OPTIONS_FILE}):\n${_options_content}\n")
+
+    # All options are listen in file, gluecodium needs only path to this file
```
limingchina, line 81: listed?

cmake/modules/gluecodium/gluecodium/Generate.cmake
```diff
@@ -170,7 +173,7 @@ function(apigen_generate)
       PRIVATE ${apigen_generate_CPP_EXPORT}_INTERNAL)
   endif()
 
-  message("Configured `Gluecodium" ${APIGEN_GLUECODIUM_ARGS})
+  message("Configured `Gluecodium")
```
limingchina, line 176: What's the extra character before "Gluecodium"?

cmake/modules/gluecodium/gluecodium/Generate.cmake
```diff
@@ -103,10 +103,10 @@ function(apigen_generate)
   _apigen_set_option_or_default(VERSION +)
 
   set(operationVerb "Generate")
-  set(validateParam "")
+  set(validateProperty "false")
   if(${apigen_generate_VALIDATE_ONLY})
     set(operationVerb "Validate")
-    set(validateParam "-validateOnly")
+    set(validateProperty "true")
```
limingchina, line 109: What does this change mean?

cmake/modules/gluecodium/gluecodium/Generate.cmake
Hsilgos: I don't know, it was done by @Dschoordsch

cmake/modules/gluecodium/gluecodium/Generate.cmake
Hsilgos: -validateOnly was command line flag to enable validation (with actual generation as i understand).
Its analog in property file is
validate=[true|false]

limingchina: Does this mean the command line option is changed? Where is the property file located and how is it specified?

cmake/modules/gluecodium/gluecodium/Generate.cmake
limingchina: Suspect it's a typo. You probably can check with him in dependent from this patch.

cmake/modules/gluecodium/gluecodium/Generate.cmake
```diff
@@ -182,50 +185,62 @@ function(apigen_generate)
...
-        -DAPIGEN_GLUECODIUM_ARGS=${APIGEN_GLUECODIUM_ARGS}
+        -DAPIGEN_GLUECODIUM_OPTIONS_FILE=${_gluecodium_options_file}
+        -DAPIGEN_GLUECODIUM_AUXINPUT_FILE=${_gluecodium_auxinput_file} # CMake < 3.15
         -DAPIGEN_GLUECODIUM_VERSION=${apigen_generate_VERSION}
         -DAPIGEN_GLUECODIUM_DIR=${APIGEN_GLUECODIUM_DIR}
         -DAPIGEN_GENERATOR=${APIGEN_GENERATOR}
         -DAPIGEN_OUTPUT_DIR=${APIGEN_OUTPUT_DIR}
         -DAPIGEN_COMMON_OUTPUT_DIR=${APIGEN_COMMON_OUTPUT_DIR}
         -DAPIGEN_BUILD_OUTPUT_DIR=${APIGEN_BUILD_OUTPUT_DIR}
-        -DAPIGEN_GRADLE_SYNCHRONISATION_DIR=${CMAKE_BINARY_DIR}
```
limingchina, line 206: Where does this go now?

Hsilgos: Good catch

cmake/modules/gluecodium/gluecodium/Generate.cmake
Hsilgos: Command line options are moved to property file. Property file is created below and it's path is specified like
> set (_gluecodium_options_file "${CMAKE_CURRENT_BINARY_DIR}/gluecodium-options-${APIGEN_TARGET}.txt")

It's unique for each source folder + target. Almost all command line options are now stored there and path to file is passed to gluecodium with only one command line option -options=<path-to-file> in runGenerate.cmake

limingchina: OK. Ic. That's property file generated later.

PR #334, opened 2020-05-11
Adds fallback patch to support obsolete CMake < 3.15

Hsilgos:

### DIFF
--- a/cmake/modules/gluecodium/gluecodium/Generate.cmake
+++ b/cmake/modules/gluecodium/gluecodium/Generate.cmake
@@ -20,7 +20,7 @@ if(DEFINED includeguard_gluecodium_Generate)
 endif()
 set(includeguard_gluecodium_Generate ON)
 
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.5)
 
 #.rst:
 # Code generation module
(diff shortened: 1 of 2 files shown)


cmake/modules/gluecodium/gluecodium/Generate.cmake
```diff
@@ -182,8 +182,15 @@ function(apigen_generate)
...
-  set (_lime_interface_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},INTERFACE_SOURCES>,INCLUDE,.*\\.lime$>")
-  set (_lime_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},SOURCES>,INCLUDE,.*\\.lime$>")
+  if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.15)
+    set (_lime_interface_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},INTERFACE_SOURCES>,INCLUDE,.*\\.lime$>")
+    set (_lime_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},SOURCES>,INCLUDE,.*\\.lime$>")
+    set (_lime_dependencies ${_lime_interface_sources} ${_lime_sources})
+  else()
+    message (WARNING "CMake < 3.15 is obsolete and may work wrongly with Gluecodium")
+    set (_lime_interface_sources "$<TARGET_PROPERTY:${APIGEN_TARGET},INTERFACE_SOURCES>")
+    set (_lime_sources "$<TARGET_PROPERTY:${APIGEN_TARGET},SOURCES>")
```
DanielKamkha, line 192: Why this branch is not setting `_lime_dependencies` var? I think it should. Or, better yet, set the var outside of the `if` below.

cmake/modules/gluecodium/gluecodium/Generate.cmake
```diff
@@ -182,8 +182,15 @@ function(apigen_generate)
...
     BUILD_OUTPUT_DIR "${APIGEN_BUILD_OUTPUT_DIR}")
 
-  set (_lime_interface_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},INTERFACE_SOURCES>,INCLUDE,.*\\.lime$>")
-  set (_lime_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},SOURCES>,INCLUDE,.*\\.lime$>")
+  if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.15)
+    set (_lime_interface_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},INTERFACE_SOURCES>,INCLUDE,.*\\.lime$>")
+    set (_lime_sources "$<FILTER:$<TARGET_PROPERTY:${APIGEN_TARGET},SOURCES>,INCLUDE,.*\\.lime$>")
+    set (_lime_dependencies ${_lime_interface_sources} ${_lime_sources})
+  else()
+    message (WARNING "CMake < 3.15 is obsolete and may work wrongly with Gluecodium")
```
DanielKamkha, line 190: "wrongly" -> "incorrectly"

cmake/modules/gluecodium/gluecodium/Generate.cmake
Hsilgos: Because for CMake  < 3.15 this variable must be empty, otherwise custom command will depend on other (than lime) sources as well, but it shouldn't

PR #378, opened 2020-06-04
Fix crash in JNI proxy cache

DanielKamkha:
Updated JNI templates for proxy cache to fix a bug in the destructor of
CppProxyBase class. Previous implementation incorrectly omitted
"type_key" string from the cache key used in the destructor, thus
failing to erase the cache entry. That led to crashes in operator==()
later when a stale entry with a deleted Java object was accessed.

Updated smoke tests. A functional regression test seems to be
infeasible. Consistent reproduction of the bug relies on the Java object
referred from the stale cache entry to be garbage-collected, and GC
action in JVM is non-enforcible and thus not consistent enough to be
reproducible in tests.


gluecodium/src/main/resources/templates/jni/utils/CppProxyBaseImplementation.mustache
```diff
@@ -39,9 +39,11 @@ CppProxyBase::getJniEnvironment( )
...
 }
 
-CppProxyBase::CppProxyBase( JNIEnv* jenv, {{>common/InternalNamespace}}jni::JniReference<jobject> globalRef, jint jHashCode )
-    : mGlobalRef( std::move( globalRef ) )
-    , jHashCode( jHashCode )
+CppProxyBase::CppProxyBase( JNIEnv* jenv,
+                            {{>common/InternalNamespace}}jni::JniReference<jobject> globalRef,
+                            jint jHashCode,
+                            ::std::string type_key )
+    : mGlobalRef( std::move( globalRef ) ), jHashCode( jHashCode ), type_key( type_key )
```
Hsilgos, line 46: std::move?

DanielKamkha: Done

gluecodium/src/main/resources/templates/jni/utils/CppProxyBaseHeader.mustache
```diff
@@ -159,6 +159,7 @@ private:
     JniReference<jobject> mGlobalRef;
     jint jHashCode;
     ReverseCacheKey mReverseCacheKey = nullptr;
+    ::std::string type_key;
```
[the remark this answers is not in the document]
DanielKamkha: Done

Issue #436, opened 2020-07-28
Add support for generating stubs in Java


DanielKamkha:
Gluecodium-generated code for structs and classes in Java is hard to mock for unit tests. The proposed solution is to add a new mode for the Java generator that would output method stubs instead of marking methods as "native". 

This option can be enabled by a new dedicated command-line parameter (e.g. "-stubs"). This way the semantics of the option could be later extended to other generators as well.

PR #444, opened 2020-07-29
Add "stubs" mode for Java generator

DanielKamkha:
Added "stubs" mode for Java generator, enabled by "-stubs" command-line
parameter. This mode replaces "native" methods in generated Java code
with stubs (i.e. empty implementation) and removes "final" qualifier
from classes and fields. This makes code generated for classes and
structs fully mockable in unit tests.

"Stubs" mode also skips generation of JNI code.

Added new smoke test group StubsSmokeTest that runs its use cases with
"stubs" mode enabled. Added smoke tests for classes, exceptions, and
structs with methods there.

Added "STUBS" Boolean parameter to apigen_generate() CMake function. Apart from
enabling "stubs" mode for Gluecodium, this flag also skip generation of
C++ code.

Resolves: #436

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Gluecodium project Release Notes
 
+## Unreleased
+### Features:
+  * Introduced "stubs" mode for Java generator (`-stubs` command-line parameter). This mode replaces
+    "native" methods in generated Java code with stubs (i.e. empty implementation) and removes
+    "final" qualifier from classes and fields. This makes code generated for classes and structs
+    fully mockable in unit tests.
+
 ## 8.0.1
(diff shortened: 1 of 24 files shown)


DanielKamkha:
> Structs look good... What about classes?

There are no "structs" in our Java templates. LimeIDL classes and LimeIDL structs are both handled by "Class" template for Java.

Hsilgos:
> > Structs look good... What about classes?
> 
> There are no "structs" in our Java templates. LimeIDL classes and LimeIDL structs are both handled by "Class" template for Java.

Yes, I know. I meant classes in lime, but I already found 'class SomeClass'. For me looks good.

Issue #446, opened 2020-07-30
Use import aliases for Dart external types


DanielKamkha:
To avoid potential name clashes with existing generated types, it would be beneficial for Dart generated code to use import aliases to import types declared as Dart-external.

PR #450, opened 2020-07-30
Use import aliases for Dart external types

DanielKamkha:
Updated DartImportResolver to add import aliases to imports for external
types. Updated DartNameResolver to use these import aliases for type
references to external types.

Added new "ref" subkey to NameResolverHelper template helper to allow
resolving type names as if they were type references.

Updated Dart templates to use "ref" subkey to resolve type names for
conversion functions.

Resolves: #446

### DIFF
--- a/gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartImportResolver.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartImportResolver.kt
@@ -60,8 +60,8 @@ internal class DartImportResolver(
             limeElement is LimeClass -> listOf(tokenCacheImport)
             else -> emptyList()
         } + listOfNotNull(
-            resolveExternalImport(limeElement, IMPORT_PATH_NAME),
-            resolveExternalImport(limeElement, CONVERTER_IMPORT_NAME)
+            resolveExternalImport(limeElement, IMPORT_PATH_NAME, useAlias = true),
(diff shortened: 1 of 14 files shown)


gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartImportResolver.kt
```diff
@@ -116,22 +119,28 @@ internal class DartImportResolver(
...
             )
         }
     }
 
     companion object {
         private val collectionSystemImport = DartImport("collection", importType = ImportType.SYSTEM)
         private val collectionPackageImport = DartImport("collection/collection")
+
+        fun computeAlias(importPath: String) =
+            importPath.split(':').last().split('/').last().split('.').first()
```
Hsilgos, line 144: May be add suffix '_' to make it 'internal'?

DanielKamkha: Import aliases are always file-private, regardless of the alias name. Optionally adding a '_' prefix (not suffix) to an alias is just pure decoration. With '_' it looks like a local private variable. Without '_' it looks like a FQN. Syntactically it is neither and the prefix does nothing.

Or did you mean something else?

Hsilgos: > Import aliases are always file-private, regardless of the alias name

Fair enough, I didn't know

Issue #458, opened 2020-08-06
Implement null safety for Dart generated code


DanielKamkha:
Dart 2.9, released on 5th of August 2020, introduces explicit null safety language feature (i.e. marking nullable type references with '?', similar to Swift and Kotlin). Dart code generated by Gluecodium needs to support that as well.

First step is to upgrade CI to use Dart 2.9 and to solve any issues that arise in the process. After that, the Dart generator needs to be updated. Both smoke and functional tests need to be added/updated as necessary.

DanielKamkha:
Null safety feature seems to be still in the "experimental" state in Dart 2.9.

PR #459, opened 2020-08-06
Upgrade Travis to use Dart 2.9

DanielKamkha:
See: #458


DanielKamkha:
Killed all caches and rebuilt from scratch, it works.

PR #475, opened 2020-08-19
Returns processing of VERBOSE flag.

Hsilgos:
Hides all output if VERBOSE is OFF


DanielKamkha (changes requested):
Please set VERBOSE to ON in Gluecodium's own functional tests. There are two directories with them: "examples" and "functional-tests".

Hsilgos:
@DanielKamkha in examples it's already ON

DanielKamkha (approved):

Issue #541, opened 2020-10-18
Failed to build example: SEVERE: Failed reading options: Unrecognized option: -nostdout


iamsergio:
Are there any requirements regarding Java and Gradle ? It fails on my Archlinux
(gradle 6.6 ; OpenJDK 1.8)

```
$ ./generate -input ${PWD}/examples/libhello/lime -output ${PWD}/generated -nostdout
Oct 18, 2020 1:15:32 PM com.here.gluecodium.Gluecodium$Companion main
INFO: Version: 8.4.5
Oct 18, 2020 1:15:33 PM com.here.gluecodium.Gluecodium$Companion main
SEVERE: Failed reading options: Unrecognized option: -nostdout
usage: generate [input] [-auxinput <arg>] [-cache] [-commonoutput <arg>]
       [-copyright <arg>] [-cppexport <arg>] [-cppnamerules <arg>]
       [-cppnamespace <arg>] [-dartlookuperrormessage <arg>]
       [-dartnamerules <arg>] [-generators <arg>] [-help] [-input <arg>]
       [-internalprefix <arg>] [-intnamespace <arg>] [-intpackage <arg>]
       [-javanamerules <arg>] [-javanonnullannotation <arg>]
       [-javanullableannotation <arg>] [-javapackage <arg>] [-libraryname
       <arg>] [-mergemanifest <arg>] [-options <arg>] [-output <arg>]
       [-stubs] [-swiftnamerules <arg>] [-validate] [-version] [-werror
       <arg>]
Gluecodium - Generate APIs from LimeIDL files

 -auxinput <arg>                                 Auxiliary IDL sources
                                                 that are loaded but not
                                                 generated from
 -cache,--enable-caching                         enable caching of output
                                                 files, only available if
                                                 output destination is set
 -commonoutput,--common-output-dir <arg>         Common generated files
                                                 output directory,
                                                 defaults to be the same
                                                 as output
 -copyright,--copyright-header <arg>             Specify the path for the
                                                 file containing the
                                                 copyright header that
                                                 will be appended to all
                                                 the generated files.
 -cppexport <arg>                                C++ export macro name for
                                                 explicit symbol
                                                 exporting.
 -cppnamerules <arg>                             C++ name rules property
                                                 file.
 -cppnamespace <arg>                             C++ namespace for public
                                                 (API) headers.
 -dartlookuperrormessage <arg>                   Custom error message for
                                                 when Dart FFI function
                                                 lookup fails.
[...]

PR #552, opened 2020-10-22
Update README files with regard to "examples" dir

DanielKamkha:
See: #541


iamsergio:
all mentions to -enableCaching can also be removed

DanielKamkha:
> all mentions to -enableCaching can also be removed

This command line parameter was renamed to -cache. Updated README to reflect that.

Hsilgos (approved):

Issue #543, opened 2020-10-20
Add support for free-form platform attributes


cuteqt:
Swift supports to add attribute to the functions. i.e. for functions with trivial return value,
without `@discardableResult` the Swift developers have to use:
_ = A.func(...)

with `@discardableResult` the user could simply use:
A.func(...)
and not worry about compiler warning.

This is just an example. There are many other useful attributes available in Swift, Java and C++. It would be nice to support it in gluecodium.

Hsilgos:
Java, Swift and C++ all support attributes on functions. It would be nice to support such attributes generally.

DanielKamkha:
Dart also supports attributes. All these languages also support attributes on most elements, not just functions.

Moreover, the existing `@Swift(ObjC)` IDL attribute could be converted to this general system.

DanielKamkha:
I propose platform-attribute syntax for this. For example, to add a `@discardableResult` attribute in Swift generated code, one would mark a function in IDL with `@Swift(Attribute="discardableResult")` IDL attribute.

DanielKamkha:
One additional possible usage for this would be platform-specific deprecation attributes, e.g. `@Java(Attribute="@Deprecated(\"foo bar\")"`. This needs to be properly documented.

PR #557, opened 2020-10-27
Add support for free-form attributes in Java generator

DanielKamkha:
Updated Java templates to add support for `@Java(Attribute="foo")` IDL attribute. This syntax should produce and
free-form `@foo` Java annotation in the generated Java code.

Added a new dedicated smoke test "attributes".

Also updated LimeAttributes.Builder class to properly support repeated attributes. The new "attributes" smoke test
contains use cases that test this support.

See: #543


gluecodium/src/test/resources/smoke/attributes/input/Attributes.lime
```diff
@@ -0,0 +1,121 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+package smoke
+
+@Java(Attribute = "OnClass")
```
Hsilgos, line 20: May be better to keep full form like `Attribute = "@OnClass" `? For example in C++ it looks like `[[attribute]]`, so may be we want to make it more explicit?

gluecodium/src/test/resources/smoke/attributes/output/android/com/example/smoke/AttributesClass.java
```diff
@@ -0,0 +1,31 @@
...
+            public void disposeNative(long handle) {
+                disposeNativeHandle(handle);
+            }
+        });
+    }
+    private static native void disposeNativeHandle(long nativeHandle);
+    @OnFunctionInClass
+    public native void veryFun(@OnParameterInClass @NonNull final String param);
+    @OnPropertyInClass
+    @NonNull
```
Hsilgos, line 27: I guess `@NonNull`, `@Deprecated`, etc can be switched to the same mechanism, it may simplify mustache files

gluecodium/src/test/resources/smoke/attributes/input/Attributes.lime
DanielKamkha: I really don't like the idea of just taking a string value "foobar" from `Attribute="foobar"` verbatim. This opens a wide door to code injection hacks from "smart" IDL authors.

Admittedly, even in the current form someone can attempt code injection, so I'll add a few guards against this right now.

gluecodium/src/test/resources/smoke/attributes/output/android/com/example/smoke/AttributesClass.java
DanielKamkha: No. Nullability in it's current form is up to standard of the modern languages (e.g. Kotlin, Swift, Dart). Shifting away from `?` syntax to ultra-verbose attribute syntax sounds bad on its own. "Simplifying templates" at the cost of forcing the user to use ridiculous amount of boilerplate is just a crime.

Similarly, current `@Deprecated` applies to all platforms automatically. Moving away from it multiplies boilerplate x4. So, no. On the bright side, the custom attributes allow for per-platform deprecation... but I need to make sure escaped quotes `\"` are processed correctly.

Hsilgos: I don't mean to change anything in lime syntax, I'm talking only about internal implementation in Gluecodium. Gluecodium may put `@Deprecated` and `@NonNull` to internal collection of attributes and mustache will have only one placeholder `{{>java/JavaAttributes}}`

DanielKamkha: Sounds doable then. But I'm still not sure it's worth it. For example, `@NonNull` is Java-specific, so it would require some additional post-processing of LIME model specifically for Java generator.

Hsilgos (approved):

Issue #582, opened 2020-11-04
Example: Add example of overriding a C++ virtual function in platform code


iamsergio:
Suppose you have a C++ class:
```
class MyClass {
    int bar() { return foo(); }
    virtual int foo() const { return 42 };
};
```

and the corresponding auto-generated Dart class too.
The user should be able to derive from MyClass in Dart, override `foo()`, and his `foo` get called whenever `bar()` gets called in C++.

Is this possible already ? If yes then an example in the examples/ directory would be great.

DanielKamkha:
It's not a scenario we've used in production or tests yet. But it looks to be possible with the current functionality. I'll commit an example and then let's see if I understood the scenario correctly.

DanielKamkha:
Added functional test as an example, please have a look: #586

Hsilgos:
I'm not sure that I understand your case correctly.
If you declare MyClass as interface Gluecodium will generate abstract class in Dart from which you may derive. Then you can pass this object to C++ and you will able to call foo and bar in C++ and corresponding method will be called in Dart.
You should try to use this mechanism.
If you want to connect non-empty manually written class in C++ with Dart (Java, Swift) directly you will have to use external types and describe all transitive types and method calls. In this case I would use delegation.

PR #586, opened 2020-11-04
Add functional test for "external" interface

DanielKamkha:
Added functional tests for using a pre-existing non-abstract C++
class as an "external" interface in IDL.

See: #582

### DIFF
--- a/examples/platforms/dart/test/ExternalTypes_test.dart
+++ b/examples/platforms/dart/test/ExternalTypes_test.dart
@@ -27,6 +27,11 @@ import "../test_suite.dart";
 
 final _testSuite = TestSuite("ExternalTypes");
 
+class MyDartClass extends MyClass {
+  @override
+  int foo() => 77;
+}
+
 void main() {
   _testSuite.test("Use external struct", () {
     final externalStruct =
@@ -93,5 +98,14 @@ void main() {
     expect(result.compressionState, struct.compressionState);
     expect(result.color, struct.color);
     expect(result.season, struct.season);
+  });
+  _testSuite.test("Use MyClass", () {
+    final useMyClass = UseMyClass();
+
+    final result = useMyClass.callBar(MyDartClass());
+
+    expect(result, 77);
+
+    useMyClass.release(); // No automatic finalizers in Dart's own FFI API yet.
   });
 }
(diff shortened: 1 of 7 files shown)


examples/platforms/dart/test/ExternalTypes_test.dart
```diff
@@ -94,4 +99,9 @@ void main() {
     expect(result.color, struct.color);
     expect(result.season, struct.season);
   });
+  _testSuite.test("Use MyClass", () {
+    final result = UseMyClass.callFoo(MyDartClass());
```
iamsergio, line 103: almost, what I meant is, here you should call bar(), and result should still be 77.
C++'s bar() then calls C++ foo(), but since foo() is virtual, and overriden in MyDartClass, the result is 77

iamsergio: also, would it be possible to have a more OOP way, like:

```
var myclass = UseMyClass();
myclass.bar();
```

without the static methods

Hsilgos: > but since foo() is virtual, and overriden in MyDartClass, the result is 77

Technically it's problematic without generation of class which derives from MyClass in C++.
Please describe a bit more your use case, probably it might be solved somehow differently.

DanielKamkha: @iamsergio, I've updated the example according to your comments.

@Hsilgos, I think you're over-complicating things. Using a pre-existing non-abstract class as an "external" interface in IDL just works, as this example clearly shows.

iamsergio:
```
class MyDartClass extends MyClass {
  @override
  int foo() => 77;
}
```

This works but forces the user to always override `foo()` , as it's not implemented in `MyClass.dart`. 
If `foo()` was pure virtual in C++ then this would make sense. Is this a technical limitation or just how things are right now ?

DanielKamkha: This is a technical limitation. Not in a sense of this being impossible, but it a sense of this being very complicated. 

Consider: even if we somehow let user to not override `foo()`, we need a way to know that a non-overridden case should call back to C++. So the full infrastructure would be "if it's overridden in Dart, call back to Dart; if it is not, call back to C++". Probably doable, but very complicated. And so far there were no such use cases in the internal projects we use Gluecodium for.

DanielKamkha: Also, constructing such object would be... interesting. Current bindings support either "Dart object backed by C++ object" or "C++ object backed by Dart object". This scenario would need a "C++ object backed by Dart object partially backed by another C++ object".

Hsilgos (approved):

Issue #603, opened 2020-11-17
Find an alternative solution for checking Dart functional tests for memory leaks


DanielKamkha:
Currently Dart functional tests are checked for memory leaks with AddressSanitizer tool. To facilitate this, we use Dart SDK recompiled to omit tcmalloc library.

Dart SDK compilation takes significant time. The whole Dart job on CI took \~40 minutes initially. At some point, for unknown reason (presumably either Dart code change or Travis server configuration change), this time grew to 50\~60 minutes.

As of today, Travis started to apply 50 minutes timeout to Gluecodium's CI jobs. Which is a correct timeout for public repositories. It is unknown why this timeout was not applied before.

Due to this new stricter timeout, Gluecodium's Dart CI job constantly fails now. So we need an alternative solution for Dart memory leak checks. Possible directions of research are:
* look for an official ready-made tcmalloc-less Dart SDK binary.
* find an alternative memory check tool that is not impeded by tcmalloc (valgrind has been tried already, does not work).
* use a docker container with pre-compiled tcmalloc-less Dart SDK (hosted where?).

DanielKamkha:
Official Dart SDK and custom-built Dart SDK seem to have different memory layouts, probably due to presence/absence of tcmalloc. Thus memchecked build needs to be ran as a separate job alongside the non-checked build on the official SDK binary.

DanielKamkha:
One possible solution is to use GitHub Actions as a CI, either for dart-mem-checked build only or for the whole Gluecodium CI. First and foremost, there are no build job timeouts imposed by GitHub actions. Moreover, there seems to be a caching solution for dockerized builds (perhaps involving a third-party action plugin).

PR #652, opened 2020-12-10
Migrate CI to GitHub Actions 

DanielKamkha:
Migrated current CI jobs from Travis to GitHub Actions.

Updated valgrind suppressions for Swift functional tests to mark new false positives on the new hardware.

Updated C++ and Swift functional test build scripts to explicitly define an error exit code for valgrind.

See: #603 


.github/workflows/gluecodium.yml
```diff
@@ -0,0 +1,207 @@
...
+          touch ~/.android/repositories.cfg
+          yes | sdkmanager --licenses
+          yes | sdkmanager "platforms;${ANDROID_API_LEVEL}" > /dev/null
+          yes | sdkmanager "build-tools;${ANDROID_BUILD_TOOLS_VERSION}" > /dev/null
+          yes | sdkmanager "cmake;${ANDROID_CMAKE_VERSION}" > /dev/null
+      - name: Build and run functional tests
+        run: |
+          export ANDROID_HOME=${HOME}/android-sdk-linux
+          export ANDROID_NDK_HOME=${ANDROID_HOME}/ndk-bundle
+          export SDK_ROOT=${ANDROID_HOME}
```
Hsilgos, line 112: I would say that SDK_ROOT or NDK_ROOT shouldn't be used

Hsilgos (approved):

.github/workflows/gluecodium.yml
DanielKamkha: Will do in the follow-up commit.

Issue #668, opened 2021-01-04
@Deprecated does not fully work for properties


DanielKamkha:
[originally reported by @OguzHere]

Adding a “Deprecated” annotation to the Section.lime file:
```
    // Describes the departure location.
    // @get Gets the departure location.
    @Deprecated("Will be removed in v4.8.0, use departurePlace.")
    property departure: Departure {
        get
    }
```

The Swift gets it:
```
    /// Describes the departure location.
    @available(*, deprecated, message: "Will be removed in v4.8.0, use departurePlace.")
    public var departure: Departure {
        get {
            return moveFromCType(sdk_routing_Section_departure_get(self.c_instance))
        }
    }
```

But, the Java hasn’t get it:
```
    /**
     * <p>Gets the departure location.</p>
     */
    @NonNull
    public native Departure getDeparture();
```

I think this is not what is expected of the @Deprecated annotation.

Flutter does’t get it, either:
```
  /// Gets the departure location.
  Departure get departure;
```

PR #669, opened 2021-01-04
Propagate `@Deprecated` attribute to property accessors

DanielKamkha:
Updated LIME model loader to propagate `@Deprecated` attribute from property itself to its accessors, but only if the
accessors lack a `@Deprecated` attribute of their own.

Added a smoke test use case for this.

Resolves: #668

### DIFF
--- a/gluecodium/src/test/resources/smoke/comments/output/lime/smoke/DeprecationComments.lime
+++ b/gluecodium/src/test/resources/smoke/comments/output/lime/smoke/DeprecationComments.lime
@@ -41,4 +41,13 @@ interface DeprecationComments {
         @Deprecated("Unfortunately, this property's setter is deprecated.\nUse [comments.SomeProperty.set] instead.")
         set
     }
+    // Describes the property but not accessors.
+    // @get Gets the property but not accessors.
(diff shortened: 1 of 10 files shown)


gluecodium/src/test/resources/smoke/comments/output/lime/smoke/DeprecationComments.lime
```diff
@@ -41,4 +41,13 @@ interface DeprecationComments {
         @Deprecated("Unfortunately, this property's setter is deprecated.\nUse [comments.SomeProperty.set] instead.")
         set
     }
+    // Describes the property but not accessors.
+    // @get Gets the property but not accessors.
+    @Deprecated("Will be removed in v3.2.1.")
+    property PropertyButNotAccessors: String {
+        @Deprecated("Will be removed in v3.2.1.")
```
Hsilgos, line 48: Shouldn't text be different for get/set to cover case when message is different?

DanielKamkha: This use case is already covered by previously existing tests.

Hsilgos (approved):

PR #680, opened 2021-01-06
Fix flaky Android functional test

DanielKamkha:
Updated ListenerWithMapsTest.checkNativeListener() Android functional test to have its own "static log" instead of
relying on global singleton HelloWorldStaticLogger. This fixes the test's flakiness.

### DIFF
--- a/examples/libhello/src/test/ListenerWithMaps.cpp
+++ b/examples/libhello/src/test/ListenerWithMaps.cpp
@@ -76,7 +76,7 @@ public:
                    << std::endl;
         }
 
-        s_log = s_log + stream.str();
+        s_log += stream.str();
     }
 };
 }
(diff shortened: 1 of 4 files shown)


examples/libhello/src/test/ListenerWithMaps.cpp
```diff
@@ -75,10 +76,10 @@ class MyForecastListener : public ::test::ForecastListener
                    << std::endl;
         }
 
-        ::hello::HelloWorldStaticLogger::append( stream.str( ) );
+        s_log = s_log + stream.str();
```
Hsilgos, line 79: s_log += stream.str();
Or even
s_log.append(stream.str());
?

DanielKamkha: Done

Hsilgos (approved):

Issue #702, opened 2021-01-18
Skip entities for tags


Hsilgos:
Currently functions, properties, etc. might be skipped depending on language using `@Java(Skip)`, `@Swift(Skip)`, `@Dart(Skip)` attributes.
As Gluecodium user I'd like to disable (skip) some functionality not only depending on language, but depending project's configuration. For example if I want versions `lite` and `professional` of my project and I have class like this:
```
class Foo {
  fun myProFunctionality()
}
```

I want to show function `myProFunctionality()` only in `professional` version and hide it for `lite`.
Currently it's not possible without ugly workarounds.
I propose to extend attributes to disable functions when one of few attributes are set. I.e. Gluecodium must have command line option `tags` which processes list of strings and Gluecodium must process new attribute `@Skip(<tag-name>)` to disable entity if given tag is specified. If few tags are specified: `@Skip(<tag-name1>), @Skip(<tag-name2>)` then Gluecodum must disable functionality if ANY of listed tags are passed with command line options. So given example should look like:
```
class Foo {
  @Skip(lite)
  fun myProFunctionality()
}
```
It's preferably to process some default tags like `Swift`, `Java`, `Dart`. So `@Java(Skip)` must be an equivalent of `@Skip(Java)`.

Optionally it would be nice if Gluecodium also would support version with list of tags like `@Skip(<tag-name1>, <tag-name2>)` which should disable entity if BOTH tags are specified. Then it could be possible to build quite complex logic like `@Skip(tag1) @Skip(tag2,tag3)` which should mean to disable entity if `tag1 OR (tag2 AND tag3)` are passed to Gluecodium.

Open question: how to generate C++ code. There are two strategies.
1. Generate C++ code regardless specified tags. In this case C++ also will be the same for any passes tags. Now Gluecodium follows this strategy with Java(Skip)/Swift(Skip)/Dart(Skip)
2. Do not generate C++ code for disabled entities. It's more suitable for the scenario.

_it's decided to follow the strategy number 2._

DanielKamkha:
I'd propose _not_ implementing the "and" logic for now. It would require some separate syntax, as currently `@Foo(Bar) @Foo(Baz)` and `@Foo(Bar, Baz)` are equivalent in LIME IDL attribute syntax.

Hsilgos:
Ah, ok, nice, then let it be like this. `@Skip(Java,Professional)` should disable entity if Java OR Professional is passed, right?

[...]

PR #704, opened 2021-01-19
Add `@Skip` attribute to LIME IDL syntax

DanielKamkha:
Updated LIME model and AntlrLimeConverter to support `@Skip` attribute. The new attribute supports both
`@Skip(CustomTag)` and `@Skip("CustomTag")` syntax. `@Skip(Java)`, `@Skip(Swift)`, and `@Skip(Dart)` are equivalent to
existing `@Java(Skip)`, `@Swift(Skip)`, and `@Dart(Skip)` respectively.

Added support for `-tag` command-line parameter for specifying a comma-separated list of custom tags. Added CMake support for this parameter.

Added smoke tests for LIME generator.

See: #702


cmake/modules/gluecodium/gluecodium/Generate.cmake
```diff
@@ -158,7 +159,7 @@ function(apigen_generate)
       DART_LIBRARY_NAME
       DART_FUNCTION_LOOKUP_ERROR_MESSAGE
       DART_NAMERULES)
-  set(multiValueArgs LIME_SOURCES FRANCA_SOURCES WERROR)
+  set(multiValueArgs LIME_SOURCES FRANCA_SOURCES WERROR TAG)
```
Hsilgos, line 162: Call it TAGS, plural, as other multi value args

DanielKamkha: Done

Hsilgos (approved):

PR #831, opened 2021-03-09
Switches namerules tests to the new cmake API

Hsilgos:


DanielKamkha (approved):

DanielKamkha (changes requested):
Fails to link on Linux.

DanielKamkha (approved):

PR #854, opened 2021-03-29
Replace "jcenter()" with "mavenCentral()" in Gradle scripts

DanielKamkha:


DanielKamkha:
Currently not possible due to https://youtrack.jetbrains.com/issue/KT-44730

Hsilgos (approved):

Issue #837, opened 2021-03-12
Add support for "optimized lists"


DanielKamkha:
There is a use case that comes up regularly: exposing a large list of structs/objects from C++ to platform side. Existing approach to lists in Gluecodium bindings results in a full copy of the list and the objects therein. This is particularly bad if the list elements are also large themselves.

One possible solution is to implement optional alternative bindings for the `List<>` type. Syntactically in IDL it can be an attribute applied to a type reference, e.g. `@Optimized`. The usage needs to be limited to read-only: only function return types and read-only properties. For struct fields it can be applied for both mutable and immutable structs. However, such structs should also carry the "taint" of being "optimized" and only usable in the same places. This needs to be validated against.

Implementation-wise, `@Optimized List<Foo>` has to be `std::vector<std::shared_ptr<Foo>>` on C++ side (which is already true for lists of classes/interfaces, but needs to be done for other types). On platform side this can be exposed as a regular `List<>` interface, backed by an internal implementation that maintains a copy of the C++ vector and fetches elements from it on demand.

The first step is to evaluate the technical feasibility of this approach.

DanielKamkha:
A prototype Java implementation demonstrated that this approach is feasible.

PR #864, opened 2021-04-16
Add "optimized lists" support in Java generator

DanielKamkha:
See: #837

### DIFF
--- a/gluecodium/src/main/resources/templates/java/AbstractNativeList.mustache
+++ b/gluecodium/src/main/resources/templates/java/AbstractNativeList.mustache
@@ -1,0 +1,191 @@
+{{!!
+  !
+  ! Copyright (C) 2016-2021 HERE Europe B.V.
+  !
+  ! Licensed under the Apache License, Version 2.0 (the "License");
+  ! you may not use this file except in compliance with the License.
+  ! You may obtain a copy of the License at
+  !
+  !     http://www.apache.org/licenses/LICENSE-2.0
+  !
+  ! Unless required by applicable law or agreed to in writing, software
+  ! distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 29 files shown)


gluecodium/src/main/resources/templates/java/AbstractNativeList.mustache
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * @exclude
+     */
+    private final class NativeIterator implements ListIterator<T> {
+
+        private int index = 0;
+
+        @Override
+        public boolean hasNext() { return index <= size() - 1; }
```
Hsilgos, line 65: Looks like it might be just return index < size();

gluecodium/src/main/resources/templates/java/AbstractNativeList.mustache
```diff
@@ -0,0 +1,205 @@
...
+
+    @Override
+    public T set(int index, T element) { throw new UnsupportedOperationException(); }
+    @Override
+    public void add(int index, T element) { throw new UnsupportedOperationException(); }
+    @Override
+    public T remove(int index) { throw new UnsupportedOperationException(); }
+
+    @Override
+    public int indexOf(Object o) {
```
Hsilgos, line 167: Potentially this and operation below might cause huge slow down without caching

DanielKamkha: Yes. These are not in the expected usage scenario though. However, might as well throw an exception here now, change to something else later if the scenario changes.

gluecodium/src/main/resources/templates/java/AbstractNativeList.mustache
DanielKamkha: Done

Hsilgos (approved):

Issue #903, opened 2021-05-17
Document Gluecodium internal architecture


DanielKamkha:
To improve developer experience for on-boarding and/or external contributors, it would be beneficial to document the internal structure of Gluecodium project.

PR #918, opened 2021-05-25
Add more Gluecodium internal documentation

DanielKamkha:
Added Gluecodium internal documentation about the testing approach and about the release process.

Resolves: #903


docs/internal/release_process.md
```diff
@@ -0,0 +1,45 @@
...
+as soon as the feature is ready.
+
+Each release raises the current Gluecodium version in a dedicated commit. The same commit should have the current
+`Unreleased` section of the changelog converted into the `Release MAJOR.MINOR.PATCH` (also have a "release date" line
+added there).
+
+After the version change is committed, a release tag needs to be created manually on the
+[Releases](https://github.com/heremaps/gluecodium/releases) page. The convention is to copy-paste the current release
+section from the changelog into the "description" field: this way the users of Gluecodium, who subscribed to release
+notifications, also get the release notes for this release.
```
Hsilgos, line 45: So after release branch is created then Appveyor pick up it automatically, right? It is worth to be mentioned. Also it would be nice to mention somewhere why GitHub actions are not used for releasing

docs/internal/testing.md
```diff
@@ -0,0 +1,46 @@
...
+
+Functional tests in Gluecodium ("functional-tests" directory) are large-scale black-box tests where the output code is
+generated based on given IDL files, compiled with the tools appropriate for each corresponding language, and then run
+through a set of unit tests which use the compiled code as a library. While the testing happens through unit tests, with
+regard to Gluecodium itself these tests are black-box tests, since they are using Gluecodium as a tool, like a user
+would.
+
+These tests could be run locally (see [Development tools](development_tools.md) for the list of required software). They
+are also run on CI (currently GitHub Actions CI, see the [config](../../.github/workflows/functional-tests.yml) for
+details).
```
Hsilgos, line 46: I would mention that generated files should also be commited and merged to show how the change influences a generated code

docs/internal/release_process.md
DanielKamkha: > Appveyor 
Did you mean GitHub Actions? If yes, then yes, it will pick it up automatically, but so it will for any other non-release branch, so it does not seems to be worth mentioning. Also, the CI picking up the release branch is not relevant to the release process as there is nothing to build or verify.

> why GitHub actions are not used for releasing
"Because no one ever suggested that before" is not the thing that should be documented. It is the thing to have a ticket created for: #919.

docs/internal/testing.md
DanielKamkha: Added a paragraph about this to "smoke tests" section.

Hsilgos (approved):

Issue #911, opened 2021-05-20
Add support for "Duration" built-in type


DanielKamkha:
In additional to the already supported `Date` IDL type, it might be beneficial to support a `Duration` type as well. It can be mapped to existing types in output languages pretty easily:
* C++: https://en.cppreference.com/w/cpp/chrono/duration
* Java: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html
* Swift: https://developer.apple.com/documentation/foundation/timeinterval
* Dart: https://api.dart.dev/stable/2.12.0/dart-core/Duration-class.html

DanielKamkha:
Preliminary proposal on C++ type is `std::chrono::time_point::duration`.

DanielKamkha:
[an additional idea from @Hsilgos]

...implement [...] Duration using external types. Put related LIME files, [as well as] Java, Dart, etc. [converters] should be shipped with Gluecodium. [...] call [it] "standard library".

DanielKamkha:
The "standard library" idea has several additional challenges, mostly related to the file system:
* where to put the "standard" IDL files and source files? Most probably Gluecodium's resources.
* how to load "standard" IDL files? Probably pre-load separately from the user model and then merge the two models just before validation.
* where to put "standard" source files in the output? Probably in the "common" dir.
* how to refer to the "standard" source files from the `external` descriptor in the "source" IDL files? Currently unknown.

DanielKamkha:
The updated proposal for handling the C++ types is to specify the precision explicitly in the IDL in a pseudo-generic manner: `Duration<Milliseconds>`, etc. Basing on the proposed use cases, the initial plan is to have only two "precisions": seconds and milliseconds. Other options could be added later if needed.

PR #946, opened 2021-06-10
Generalize LimeModelFilter into LimeModelTransformer

DanielKamkha:
Generalized LimeModelFilter from a tree walker taking a `(LimeNamedElement) -> Boolean` predicate into the
LimeModelTransformer that takes a `(LimeNamedElement) -> LimeNamedElement?` transformation function.

This allows reusing the same tree walking logic for "patching" the model tree. This will be used for replacing upcoming
`Duration` types with `typealias TimeInterval = Double` to enable correct overloads resolution while generating Swift
code.

See: #911


DanielKamkha:
Cancelled.

PR #947, opened 2021-06-11
Collate Swift generics with correct name resolver

DanielKamkha:
Updated Swift generator to use Swift name resolver when collating generics, instead of CBridge name resolver as before.


DanielKamkha:
Cancelled.

PR #949, opened 2021-06-16
Fix platform filtering for functional tests

DanielKamkha:
Updated functional tests CMake scripts and build scripts to correctly filter functional test "features" (as listed in
"functional/CMakeLists.txt") per platform.

This functionality was inadverently broken during the switch to the new CMake infrastructure. All enabled generators
were taken into account, resulting in no filtering, as "cpp" generator is always present, thus enabling all tests for
all platforms. The updated scripts correctly ignore "cpp" generator when other generators are present.


DanielKamkha:
Superseded by https://github.com/heremaps/gluecodium/pull/963

PR #968, opened 2021-06-24
Use "dart pub get" instead of "pub get" for Dart functional tests

DanielKamkha:
"pub get" is a deprecated way to get Dart dependencies. Updated
CMake script for Dart functional tests to use "dart pub get" instead.


DanielKamkha:
Not needed.

Issue #550, opened 2020-10-21
Add enum to string conversion in C++


lorensi:
It's quite common for developers to create their own conversion function. If these conversions could be created by Gluecodium itself that would eliminate a bunch of user code.  

lime:
```
package sdk;

enum Foo {
  BAR,
  BAZ
}
```

Possible generated code:
```cpp
namespace sdk {

enum class Foo {
  BAR,
  BAZ
}

std::string_view                    // or std::optional<std::string_view>
to_string(Foo foo) {
  switch (foo) {
    case Foo::BAR: return "BAR"sv;  // or "Foo::Bar"
    case Foo::BAZ: return "BAZ"sv;  // or "Foo::Baz"
  }
  return "<unknown>"sv;             // or std::nullopt; 
}

}
```

Usage:
```cpp
#include <sdk/Foo.h>

namespace something {
  void compute(Foo foo) {
     std::cout << to_string(foo); // Note the unqualified call. ADL taking place here.
  }
}
```

If it's a problem to create those functions for all enums, I suggest then to create at least a new attribute to indicate that the user wants these functions to be generated by Gluecodium, like the following:

```
@Cpp(Stringify)
enum Foo {
  BAR,
  BAZ
}
```

DanielKamkha:
Creating a new attribute makes sense. Generating this for every enum would have been too much dead code.

Hsilgos:
If such generated code will be used only internally (not through external functions) then those converters might be header-only and compiler won't add them to final binary (except those which are really were used)

PR #974, opened 2021-06-30
Add to_string helper for C++ enums (#550)

piotrkania-here:
Resolves: #550
Signed-off-by: Piotr Kania <ext-piotr.kania@here.com>


gluecodium/src/main/resources/templates/cpp/CppEnumeration.mustache
```diff
@@ -26,4 +26,16 @@ enum class {{>cpp/CppAttributesInline}}{{resolveName}} {
 }}    {{resolveName}}{{#explicitValue}} = {{resolveName}}{{/explicitValue}}{{#if iter.hasNext}},{{/if}}
 {{/enumerators}}
 };
+
+{{#if attributes.cpp.tostring}}
+std::string_view
+to_string({{resolveName}} foo) {
```
DanielKamkha, line 32: Please rename `foo` to something more meaningful.

gluecodium/src/main/resources/templates/cpp/CppEnumeration.mustache
```diff
@@ -26,4 +26,16 @@ enum class {{>cpp/CppAttributesInline}}{{resolveName}} {
...
+  switch (foo) {
+    {{#enumerators}}
+    case {{resolveName}}{{explicitValue}}: return "{{resolveName}}{{explicitValue}}";
```
DanielKamkha, line 35: This will not compile. Name of the enum type is missing, `::` is missing, `explicitValue` is not needed at all. I think the correct template is something like
```
    {{#set enum=this}}{{#enumerators}}
    case {{resolveName enum}}::{{resolveName}}: return "{{resolveName enum}}::{{resolveName}}";
    {{/enumerators}}{{/set}}
```

DanielKamkha (changes requested):
Please add a functional test that compiles the generated code with the new feature and also does some simple runtime checks for the returned string value.

Please document the new feature in "lime_idl.md". Please add a changelog entry to "CHANGELOG.md"

functional-tests/functional/input/lime/Enums.lime
```diff
@@ -62,3 +62,9 @@ enum EnumWithDeprecatedItems {
     bar,
     baz
 }
+
+@Cpp(ToString)
```
Hsilgos, line 66: Since function is header only it doesn't make sense to introduce new attribute. Or move implementation to cpp.

gluecodium/src/main/resources/templates/cpp/CppEnumeration.mustache
```diff
@@ -26,4 +26,16 @@ enum class {{>cpp/CppAttributesInline}}{{resolveName}} {
 }}    {{resolveName}}{{#explicitValue}} = {{resolveName}}{{/explicitValue}}{{#if iter.hasNext}},{{/if}}
 {{/enumerators}}
 };
+
+{{#if attributes.cpp.tostring}}
+std::string
```
Hsilgos, line 31: I'd say we may use std::string_view here

functional-tests/functional/input/lime/Enums.lime
piotrkania-here: I've seen your comment in the ticket that compiler will leave out the function if not used and probably should defend my implementation on my own - but this is Daniel's decision.

DanielKamkha: As we have C++ public APIs now, this needs to be guarded by an attribute. We are not going to pollute the whole public API with this new function all of a sudden.

DanielKamkha (approved):

PR #979, opened 2021-07-01
Document skip/enable attributes precedence

DanielKamkha:


saebel1 (approved):
Looks good to me.

Issue #895, opened 2021-05-06
Fix issues from flutter analyze


Hsilgos:
`flutter analyze` shows huge amount of issued. 85% of them are `non_constant_identifier_names`. I suspect it doesn't like variable `__lib`
5% are about unused imports. 
Notes should be fixed if possible or ignored during `flutter analyze`

DanielKamkha:
One group of the issues that cannot be addressed is lower_snake_case for parameter names of `fromLambdas()` factory constructor for interfaces. These are named parameters and the constructor is public, so renaming the parameters is a breaking change of the public API.

There is, however, a change planned where this constructor is renamed and the parameters are made positional. That'd be a good opportunity to rename the parameters to lowerCamelCase as well.

hitman-codehq:
There are a few more warnings to be addressed, so I'll add each of them as a separate comment in the following format:

info • Unnecessary cast • lib/src/sdk/my_modules/my_class.dart:387:24 • unnecessary_cast

This happens due to the following code fragment:

MyClass$Impl {
  @override
  bool operator ==(dynamic other) {
    if (identical(this, other)) return true;
    if (other is! MyClass$Impl) return false;
    // This causes the warning.  It is already known that this is a MyClass$Impl so there is no need to cast it here
    return __areEqual((this as MyClass$Impl).handle, other.handle) != 0;
  }
}

hitman-codehq:
info • Unnecessary cast • lib/src/sdk/my_modules/my_class.dart:387:24 • unnecessary_cast

This happens due to the following code fragment:

```
MyClass MyClassFromFfi(Pointer<Void> handle) {
  final instance = __lib.getCachedInstance(handle);
  // This causes the warning.  It is already known that this is a MyClass so there is no need to cast it here
  if (instance != null && instance is MyClass) return instance as MyClass;
  ...
}
```

hitman-codehq:
info • Dead code • lib/src/sdk/my_modules/my_class/my_type.dart:55:3 • dead_code

This happens due to the following code fragment:

```
int sdkMyClassMyTypeToFfi(MyType value) {
  switch (value) {
  case XXX:
    return 0;
  // This causes the warning.  The break isn't reached due to the return
  break;
  case YYY:
    return 1;
  // This causes the warning.  The break isn't reached due to the return
  break;
  default:
    throw StateError("Invalid enum value $value for MyType enum.");
  }
}
```
[...]

PR #999, opened 2021-07-20
Fix another batch of "dart analyze" warnings

DanielKamkha:
Moved the suppression comment for "deprecated_member_use_from_same_package" Dart analyzer warnings
to the corrent line for it to have effect.

Removed extraneous "dart:collections" import for `@Equatable` structs with collection fields.

Resolves: #895


DanielKamkha:
Not needed, need to be suppressed through analysis_options.yaml instead.

Issue #996, opened 2021-07-20
Add possibility to enable for language with tag in opposite to skipping


Hsilgos:
Gluecodium has possibility to skip or enable like `@Skip([Tag =] "CustomTag")`/`@EnableIf([Tag =] "CustomTag")` and Gluecodium has possibility to skip for language like `@Dart(Skip = "CustomTag")` and it would be nice to align and have possibility to enable for language like `@Dart(EnableIf="CustomTag")`.

PR #1024, opened 2021-07-30
Add support for in-platform tag-enabling

DanielKamkha:
Updated LimeModelSkipPredicates to support `@Java(EnableIf = "CustomTag")` syntax (also for Swift
and Dart). Updating the predicates is sufficient to add the support for all three platforms, no
changes in any individual generator are required.

Resolves: #996

### DIFF
--- a/lime-runtime/src/main/java/com/here/gluecodium/common/LimeModelSkipPredicates.kt
+++ b/lime-runtime/src/main/java/com/here/gluecodium/common/LimeModelSkipPredicates.kt
@@ -21,6 +21,7 @@ package com.here.gluecodium.common
 
 import com.here.gluecodium.model.lime.LimeAttributeType
 import com.here.gluecodium.model.lime.LimeAttributeValueType
+import com.here.gluecodium.model.lime.LimeAttributeValueType.ENABLE_IF
 import com.here.gluecodium.model.lime.LimeAttributeValueType.SKIP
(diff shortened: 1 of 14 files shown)


lime-runtime/src/main/java/com/here/gluecodium/common/LimeModelSkipPredicates.kt
```diff
@@ -64,6 +65,18 @@ object LimeModelSkipPredicates {
...
+    private fun isSkippedByTagsOnPlatform(
+        limeElement: LimeNamedElement,
+        activeTags: Set<String>,
+        platformAttribute: LimeAttributeType
+    ): Boolean {
+        val isEnabled = hasTagsMatch(limeElement, platformAttribute, ENABLE_IF, activeTags)
+        if (isEnabled == false) return true
+
+        val isSkipped = hasTagsMatch(limeElement, platformAttribute, SKIP, activeTags)
+        return isSkipped ?: false
```
Hsilgos, line 77: Just `return isSkipped` ?

DanielKamkha: `hasTagsMatch()` returns a `Boolean?`, not `Boolean`. So neither of your suggestions will work.

Hsilgos (approved):

Issue #1052, opened 2021-08-23
Add possibility for fine-grained control over field-based struct constructors


DanielKamkha:
[requested by @jelemar]

Currently if a struct type does not have any custom constructors, a set of field-based constructors is generated for the structs. The IDL author currently has very little control over which constructors are generated. The IDL author also has no means of deprecating any of these constructors or of representing "old" field constructors when fields are added/removed.

The provisional proposed syntax is
```
field constructor(<list of field names>)
```

If no `field constructor` is specified for the struct, it should behave as it does now: having "automatic" field constructors if no "custom" constructors are present, or having "custom" constructors only if they are present.

The new "field constructors" should satisfy these requirements, if feasible:
* Attributes (skips, deprecations, etc.) and doc comments should work on them.
* Specifically for Dart, a `@Dart("name")` attribute should work correctly.
* Omitting fields from the parameter list should be only possible for fields with default values. Omitting a non-defaulted field should result in a validation error.
* A mix of "field constructors" and "custom" constructors should be possible. Constructor overloads validation should work in this scenario, unless proven too complex.

DanielKamkha:
`@immutable` structs need to have an all-fields constructor, if no "custom" constructors are present.

DanielKamkha:
Field constructors is also the first element type that can be skippable in C++. Therefore a support for `@Cpp(Skip)`  attribute is needed. Such attribute needs to be validated to be only applicable to field constructors. Probably should be implemented separately afterwards.

PR #1067, opened 2021-09-06
Add validation for field constructors overloads

DanielKamkha:
See: #1052

### DIFF
--- a/gluecodium/src/main/java/com/here/gluecodium/validator/LimeOverloadsValidator.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/validator/LimeOverloadsValidator.kt
@@ -21,17 +21,19 @@ package com.here.gluecodium.validator
 
 import com.here.gluecodium.common.LimeLogger
 import com.here.gluecodium.model.lime.LimeElement
+import com.here.gluecodium.model.lime.LimeFieldConstructor
 import com.here.gluecodium.model.lime.LimeFunction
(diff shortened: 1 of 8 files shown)


gluecodium/src/main/java/com/here/gluecodium/validator/LimeOverloadsValidator.kt
```diff
@@ -41,10 +43,29 @@ internal class LimeOverloadsValidator(
...
+    private fun validateFieldConstructor(
+        limeFieldConstructor: LimeFieldConstructor,
+        signatureResolver: LimeSignatureResolver
+    ): Boolean {
+        if (validateCustomConstructors) {
+            if (signatureResolver.hasConstructorSignatureClash(limeFieldConstructor.asFunction())) {
+                logger.error(limeFieldConstructor, "field constructor has conflicting overloads")
+                return false
+            }
+        } else {
```
piotrkania-here, line 62: Maybe we still should validate non custom constructors?
Or does heaving custom ones means there is no defaults?

DanielKamkha: If either "field" or "custom" constructors are present, there is no implicitly generated constructors anymore. So no signature/name clash with those would be possible.

piotrkania-here (approved):

PR #1098, opened 2021-09-28
Add support for multiple inheritance in Java

DanielKamkha:
Updated Java templates for classes and interfaces to support multiple
inheritance.

Updated JNI template for instance conversion to force proxy creation for narrow
interfaces (instead of trying to extract a native handle). This circumvents the
potential issues of invalid reinterpret-casts for second+ parent types. This
behavior is already documented in "lime_idl.md".

Resolves: #1079

### DIFF
--- a/functional-tests/functional/android/src/test/java/com/here/android/test/MultipleInheritanceTest.java
+++ b/functional-tests/functional/android/src/test/java/com/here/android/test/MultipleInheritanceTest.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright (C) 2016-2021 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+package com.here.android.test;
+
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertTrue;
+
+import android.os.Build;
+import com.here.android.RobolectricApplication;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+@RunWith(RobolectricTestRunner.class)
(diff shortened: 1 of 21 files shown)


functional-tests/functional/android/src/test/java/com/here/android/test/MultipleInheritanceTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    boolean result = MultipleInheritanceChecker.checkSingletonEquality(instance);
+
+    assertFalse(result);
+  }
+
+  @Test
+  public void fromCppRoundTripWithUpcastNotEquals() {
+    MultiClass uncastInstance = MultipleInheritanceFactory.getMultiClass();
+    NarrowInterface instance = (NarrowInterface)uncastInstance;
```
Hsilgos, line 112: Shouldn't it work without explicit cast?

functional-tests/functional/android/src/test/java/com/here/android/test/MultipleInheritanceTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    boolean result = MultipleInheritanceChecker.checkNarrowEquality(instance, instance);
+
+    assertTrue(result);
+  }
+
+  @Test
+  public void fromJavaRoundTripEquals() {
+    MultiInterface uncastInstance = new MultiInterfaceImpl();
+    NarrowInterface instance = (NarrowInterface)uncastInstance;
```
Hsilgos, line 150: The same

functional-tests/functional/input/src/cpp/MultipleInheritance.cpp
```diff
@@ -77,4 +80,47 @@ MultipleInheritanceFactory::get_multi_interface() {
...
+}
+
+std::shared_ptr<NarrowInterface>
+MultipleInheritanceFactory::get_multi_class_singleton() {
+    return s_multi_class;
+}
+
+std::shared_ptr<NarrowInterface>
+MultipleInheritanceFactory::upcast_multi_interface_to_narrow(const std::shared_ptr<MultiInterface>& instance) {
+    return std::dynamic_pointer_cast<NarrowInterface>(instance);
```
Hsilgos, line 95: Same as in Java, should work implicitly  here and below

functional-tests/functional/android/src/test/java/com/here/android/test/MultipleInheritanceTest.java
DanielKamkha: Yes, it will work. The explicit cast just highlights the cast itself, because it is important for the test scenario.

functional-tests/functional/android/src/test/java/com/here/android/test/MultipleInheritanceTest.java
DanielKamkha: Same as above.

functional-tests/functional/input/src/cpp/MultipleInheritance.cpp
DanielKamkha: Done

Hsilgos (approved):

PR #1099, opened 2021-09-28
Fix deadlock in Dart finalizers

DanielKamkha:
Updated FFI instance cache template to limit mutex lock lifetime to an absolute
minimum. Specifically, the new version avoids calling any Dart_DL API functions
while holding the lock. Some (or all) of these functions may cause a finalizer
to run synchronously, thus creating a double-acquire deadlock. The new version
of the code fixes this issue.

### DIFF
--- a/gluecodium/src/main/resources/templates/ffi/FfiInstanceCacheImpl.mustache
+++ b/gluecodium/src/main/resources/templates/ffi/FfiInstanceCacheImpl.mustache
@@ -35,6 +35,31 @@ std::unordered_map<InstanceCacheKey, Dart_WeakPersistentHandle> _instance_cache{
 std::mutex _instance_cache_mutex{};
 
 std::unordered_map<int32_t, std::vector<Dart_WeakPersistentHandle>> _deletion_queue{};
+
+std::vector<Dart_WeakPersistentHandle>
+{{libraryName}}_store_handle_and_get_deletion_queue_locked(
+    void* raw_pointer, int32_t isolate_id, Dart_WeakPersistentHandle new_handle
+) {
+    const std::lock_guard<std::mutex> lock(_instance_cache_mutex);
+    _instance_cache[raw_pointer] = new_handle;
+
+    auto iter = _deletion_queue.find(isolate_id);
+    if (iter == _deletion_queue.end()) return {};
+
+    auto result = std::move(iter->second);
+    _deletion_queue.erase(iter);
+    return result;
+}
+
+Dart_WeakPersistentHandle
+{{libraryName}}_get_handle_locked(void* raw_pointer)
+{
+    const std::lock_guard<std::mutex> lock(_instance_cache_mutex);
+    auto iter = _instance_cache.find(raw_pointer);
+    return (iter != _instance_cache.end()) ? iter->second : nullptr;
+}
+
+
 }
 
 #ifdef __cplusplus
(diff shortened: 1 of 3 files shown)


gluecodium/src/main/resources/templates/ffi/FfiInstanceCacheImpl.mustache
```diff
@@ -49,23 +49,27 @@ void*
...
-        Dart_DeleteWeakPersistentHandle_DL(weak_handle);
+    {
+        const std::lock_guard<std::mutex> lock(_instance_cache_mutex);
+        _instance_cache[raw_pointer] = new_handle;
+
+        // Copy deletion queue
+        auto iter = _deletion_queue.find(isolate_id);
+        if (iter == _deletion_queue.end()) return;
+
+        pending_for_deletion = iter->second;
```
Hsilgos, line 65: std swap or std move is better

gluecodium/src/test/resources/smoke/equatable/output/dart/ffi/InstanceCache.cpp
```diff
@@ -20,21 +20,23 @@ void*
...
+library_cache_dart_handle_by_raw_pointer(void* raw_pointer, int32_t isolate_id, Dart_Handle dart_handle) {
+    auto new_handle = Dart_NewWeakPersistentHandle_DL(dart_handle, NULL, 0, &library_nop_finalizer);
+    std::vector<Dart_WeakPersistentHandle> pending_for_deletion;
+    {
+        const std::lock_guard<std::mutex> lock(_instance_cache_mutex);
+        _instance_cache[raw_pointer] = new_handle;
+        // Copy deletion queue
+        auto iter = _deletion_queue.find(isolate_id);
+        if (iter == _deletion_queue.end()) return;
+        pending_for_deletion = iter->second;
```
Hsilgos, line 33: The same

gluecodium/src/main/resources/templates/ffi/FfiInstanceCacheImpl.mustache
Hsilgos: Std swap here should be more efficient. But most efficient is to extract all block to separate function and use std move inside to avoid unnecessary construction of vector.

DanielKamkha: I like the "function" approach the best, as it makes uses of RVO and thus does not need an explicit move.

gluecodium/src/test/resources/smoke/equatable/output/dart/ffi/InstanceCache.cpp
DanielKamkha: The same

gluecodium/src/main/resources/templates/ffi/FfiInstanceCacheImpl.mustache
DanielKamkha: Ah, no, still need a move here. But not below.

DanielKamkha: After some experimentation, `std::swap()` has proven to be more suitable for the vector. Preserving an empty vector in the cache, for reuse, fits the use case very well.

gluecodium/src/main/resources/templates/ffi/FfiInstanceCacheImpl.mustache
```diff
@@ -35,6 +35,31 @@ std::unordered_map<InstanceCacheKey, Dart_WeakPersistentHandle> _instance_cache{
...
+) {
+    const std::lock_guard<std::mutex> lock(_instance_cache_mutex);
+    _instance_cache[raw_pointer] = new_handle;
+
+    auto iter = _deletion_queue.find(isolate_id);
+    if (iter == _deletion_queue.end()) return {};
+
+    std::vector<Dart_WeakPersistentHandle> result;
+    result.swap(iter->second);
+    return result;
```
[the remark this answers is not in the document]
DanielKamkha: Done

Hsilgos (approved):

Issue #1190, opened 2021-11-29
Use custom Duration type in Java


DanielKamkha:
Built-in Java `Duration` type is not available in Android before API level 29. For backward compatibility, we need to generate our own custom `Duration` type. The custom type should have the same API as the standard one.

PR #1191, opened 2021-11-30
Use custom Duration type in Java

DanielKamkha:
Resolves: #1190


gluecodium/src/main/resources/templates/java/Duration.mustache
```diff
@@ -0,0 +1,367 @@
...
+        return mSeconds;
+    }
+
+    /**
+     * Gets the seconds part of this duration.
+     *
+     * @return The seconds part of this duration, value from 0 to 59.
+     */
+    public int toSecondsPart() {
+        return (int) (mSeconds % 60);
```
Hsilgos, line 272: Mixing of magic numbers and constants is not perfect

Hsilgos (approved):

Issue #1333, opened 2022-05-20
Phase out `internal` visibility


DanielKamkha:
`internal` visibility in LIME IDL was introduced years ago as a tool to prevent exposure of interim/WIP features to the public APIs while they are being worked on. It never worked completely, with only Swift supporting exact "internal" semantics. Java generator uses "package-private" visibility as "internal", and C++ just ignores `internal` completely. 

With the introduction of Dart generator, the situation worsened considerably: as Dart language only has "public" and "file-private" as visibility level, `internal` became "public" in Dart as well. So inefficient attempts to combat that were made, but in the end Dart language design prevailed.

Overall, the current state of `internal` is a large maintenance burden on Gluecodium code itself, and also a source of unending headaches for our internal projects that use Gluecodium to generate public APIs.

Currently, in out internal projects, there are three major use cases for `internal`:
* really "internal", a (small) generated API surface that is exposed to be used by some bits of hand-written code in Java or Swift.
* work-in-progress, but already used in C++ production code.
* work-in-progress, but not even used in C++, outside of non-production "experimental" builds.

Accordingly, the proposal is to replace each of these three differently:
* for the "real" internal usage, introduce an `@Internal` attribute, de-emphasizing the feature and, ideally, restricting its usage to type-level only (i.e. no more internal functions, internal fields, etc.).
* for the "WIP but we use it in C++" case, `@Skip(Java, Swift, Dart)` is a perfect solution that can be applied immediately, with no Gluecodium changes needed.
* for the "WIP experimental" case, `@EnableIf("MyExperimentalTag")` can be used, with appropriate build-system-level tags, also with no Gluecodium changes.

With this approach, `internal` visibility usage can be phase out and, eventually, `internal` support removed from Gluecodium completely.

DanielKamkha:
In the end, this proves to be infeasible. The only real change coming out if this is reducing leakage of "internal" things in Dart:
* make "internal" fields file-private in Dart
* auto-skip "internal" functions and properties in Dart

PR #1348, opened 2022-05-30
Add preliminary support for `@Internal` attribute

DanielKamkha:
Added support for `@Internal` attributes. Currently this attribute has the same
effect as `internal` visibility, with a limitation that the attribute only work
on type declarations (i.e. it doesn't work on functions, properties,
fields, and constants).

The attribute could be used interchangeably instead of- or redundantly
together with `internal` visibility. This is intended to help a gradual
phase-out of the visibility syntax.

See: #1333


DanielKamkha:
Needs some Dart support as well (with semantics of excluding the type from exports).

Hsilgos (dismissed):

Hsilgos (approved):

Issue #1355, opened 2022-05-31
Add support for "constant references"


DanielKamkha:
A use case has come up in our internal API research where it would be useful to have a LIME IDL definition that initializes a struct field with some LIME-defined constant. This is currently not supported, as the LIME model does not have a notion of a "reference to constant".

The templates for the output languages need to support this as well.

PR #1361, opened 2022-06-02
Add missing includes/imports for struct initializers and enumerator refs

DanielKamkha:
See: #1355


DanielKamkha:
Needs to be expanded to other generators as well. Needs more tests: cross-package enums, external enums.

PR #1379, opened 2022-06-16
Always skip internal functions in Dart

DanielKamkha:
Updated Dart generator and templates to always skip "internal" functions,
properties, constructors, and field constructors. That prevents accidental leaks
of internal APIs to public (the leaks happen due to lack of real "internal"
visibility in Dart).

See: #1333


DanielKamkha:
Breaking change. Postponing to combine with other upcoming breaking changes.

PR #1381, opened 2022-06-16
Update guide.md

abhiditi:

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -111,6 +111,7 @@ types can be added in the form `[package.Interface.method]`. For properties the 
 explicitly. Without suffix the reference will link to the getter in languages which have
 getters and setters.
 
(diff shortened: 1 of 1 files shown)


docs/guide.md
```diff
@@ -111,6 +111,7 @@ types can be added in the form `[package.Interface.method]`. For properties the
 explicitly. Without suffix the reference will link to the getter in languages which have
 getters and setters.
 
+**Note:** Gluecodium supports Markdown tables in Lime files. However, by default, tools like Jazzy, Dartdoc and Javadoc render these tables with invisible borders and without any padding. Therefore, we recommend to avoid using tables.
```
DanielKamkha, line 114: Generally, Gluecodium documentation does not advise on any usage patterns of its features. 

Moreover, we aim to keep the documentation as tool-agnostic as possible. Therefore, documenting quirks of third-party tools is out of the scope of this documentation.

While this sounds like a sensible piece of advice, please place it where it belongs: into the internal documentation of your project which uses all four tools (Gluecodium and the doc tools) together.

DanielKamkha (changes requested):

PR #1395, opened 2022-06-23
Rework C++ callbacks for `@Async`

DanielKamkha:
Rework C++ support for `@Async` functions to have two separate callbacks for
"result" and "error". This simplifies the usage and removes a confusing Boolean
flag from the signature.

Updated Dart bindings accordingly. Dart API surface does not change.


docs/async.md
```diff
@@ -52,21 +52,17 @@ This LIME IDL definition will result in the following C++ code being generated:
...
-* `fun myAsyncFunction(...) throws MyException` -> `std::function<void(bool /*has_value*/, MyException /*error*/)>`
-* `fun myAsyncFunction(...): ReturnType throws MyException` -> `std::function<void(bool /*has_value*/, ReturnType /*result*/, MyException /*error*/)>`
-
-All parameters passed to the "completer callback" are transformed into the appropriate state of the `Future` object on
-Dart side, allowing for Dart-idiomatic asynchronous usage.
+value. Instead:
+* The return value should be passed to the "result callback".
+* If the return type is `Void`, the "result callback" still needs to be called.
+* If the function has a `throws` clause, the error value, if present, should be passed to the "error callback".
+* Calling *both* callbacks in the same code path is invalid and leads to an undefined behavior.
```
Hsilgos, line 65: I think that behavior is quite defined.

> Calling [complete](https://api.flutter.dev/flutter/dart-async/Completer/complete.html) or [completeError](https://api.flutter.dev/flutter/dart-async/Completer/completeError.html) must be done at most once.

Flutter will fail with error "Future already complete", for calling C++ it most probably means the same hang is I encountered previously

DanielKamkha: Updated.

Hsilgos (approved):

Issue #1422, opened 2022-07-21
Phase out support for C++11


DanielKamkha:
For historic reasons, Gluecodium still support C++11. Meaning that any C++14 and C++17 features are either not used or used through some complex drop-in replacements.

Considering that C++11 is now more than 10 years old, the proposal is to switch to C++17 as the minimum version now. That would involve several steps:
* upgrade Gluecodium's functional tests for all platforms to use C++17
* remove C++ functional tests for C++11 and C++14 (there are some)
* remove the drop-in replacements for C++17 types, while keeping the typedefs
* remove the typedefs

PR #1427, opened 2022-07-22
Remove the custom "optional" type

DanielKamkha:
See: #1422


gluecodium/src/main/resources/templates/cpp/common/Optional.mustache
```diff
@@ -65,16 +65,11 @@
 #endif
 
 #if !defined(GLUECODIUM_OPTIONAL_HEADER) && !defined(GLUECODIUM_OPTIONAL_TYPE)
```
Hsilgos, line 67: Description above should be changed slightly, this should be removed

> fallback to a built-in optional implementation otherwise.

DanielKamkha: Done

Hsilgos (approved):

Issue #1423, opened 2022-07-21
Use std::string_view instead of std::string for keys in Java class cache 


Hsilgos:
Current implementation looks like
```
std::unordered_map<std::string, jni::JniReference<jclass>*>& get_instance_class_map()
{
    static std::unordered_map<std::string, jni::JniReference<jclass>*> classes;
    return classes;
}
```
This map is basically filled once at start and then immutable. Makes sense to use std::string_view as key after Gluecodium switches to C++17. 
Consider to use std::string_view for `TypeRepository`, `CachedJavaClass`, etc.

DanielKamkha:
Prerequisite: #1422

PR #1429, opened 2022-07-26
Use `std::string_view` as a type id in the TypeRepository

DanielKamkha:
Resolves:  #1423

### DIFF
--- a/gluecodium/src/main/resources/templates/jni/utils/JniClassCacheHeader.mustache
+++ b/gluecodium/src/main/resources/templates/jni/utils/JniClassCacheHeader.mustache
@@ -23,6 +23,7 @@
 #pragma once
 
 #include "JniReference.h"
+#include <string_view>
 
 {{#internalNamespace}}
 namespace {{.}}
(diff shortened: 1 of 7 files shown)


gluecodium/src/main/resources/templates/jni/utils/JniClassCacheHeader.mustache
```diff
@@ -41,7 +44,7 @@ public:
     static void init(JNIEnv* env);
 
 protected:
-    static jni::JniReference<jclass>* get_java_class_for_instance(const std::string& id);
+    static jni::JniReference<jclass>* get_java_class_for_instance(const TypeId id);
```
Hsilgos, line 47: Fields of this class (m_name, m_cpp_name) and constructor parameters may also be std::string_view now (in separate commit?)

Hsilgos (dismissed):

Hsilgos (approved):

gluecodium/src/main/resources/templates/jni/utils/JniClassCacheHeader.mustache
DanielKamkha: There is no benefit to it. These fields are initialized with static string literals. I would even expect using `string_view` would take more memory than just a `const char*`.

PR #1448, opened 2022-08-09
Document generated struct constructors

DanielKamkha:

### DIFF
--- a/docs/struct_constructors.md
+++ b/docs/struct_constructors.md
@@ -1,0 +1,131 @@
+Struct constructors
+===================
+
+This documents various types of constructors that can be defined for structs in LIME IDL.
+
+Implicit constructors
+---------------------
+
+When no constructors are explicitly defined for a LIME IDL struct, Gluecodium still adds some constructors to the
+generated code. Each such struct would have at least the constructor that provides values to all uninitialized fields 
+(i.e. fields which do not have a default value specified in the LIME definition). The order of constructor parameters
+corresponds to the order of field definitions.
+
+For example,
+```
+struct MyStruct {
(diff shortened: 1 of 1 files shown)


docs/struct_constructors.md
```diff
@@ -0,0 +1,118 @@
+Struct constructors
+===================
+
+This documents various types of constructors that can be defined for structs in LIME IDL.
+
+Implicit constructors
```
Hsilgos, line 6: I would add that implicit and field constructors are generated in specific language only without crossing languages boundaries.
Also despite that it's quite obvious for implicit constructors I would specify order in which parameters appear in constructor

docs/struct_constructors.md
```diff
@@ -0,0 +1,118 @@
...
+    isMultiline: Boolean = true
+}
+```
+will have both `MyStruct(text)` and `MyStruct(text, isMultiline)` constructors generated.
+
+Implicit constructors are sufficient for most use cases. When the API semantics require a different set of constructors,
+or when constructors need to have logic beyond just field initialization, explicit constructor definitions could be used
+instead, as outlined in the following sections.
+
+Field constructors
```
Hsilgos, line 34: Is definition of field and/or custom constructor preserves implicit ones?

docs/struct_constructors.md
DanielKamkha: Updated

docs/struct_constructors.md
DanielKamkha: Updated

Hsilgos (approved):

PR #1449, opened 2022-08-09
Add detailed documentation for `external` descriptors

DanielKamkha:


Hsilgos (approved):
I would also add typical usages like "Use native color representation" and "Store and access android Context"

PR #1457, opened 2022-08-12
Reduce visibility propagation

DanielKamkha:
Removed the logic for propagating visibility from outer to inner elements from
the AntlrLimeModel builder. For most languages, marking the inner elements
"internal" is redundant if the outer element is already marked.

Added visibility propagation on case-by-case basis where needed:
* in Dart for nested types, as there is no nesting in Dart
* in Swift for types nested inside interfaces, as there is no nesting inside
Swift protocols

For both cases, the propagation affects types only. Non-type elements are not
affected.

This both simplifies the visibility logic, and ensures that "global" and
"per-platform" visibilities are handled in the same way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Gluecodium project Release Notes
 
+## Unreleased
+### Bug fixes:
+  * Removed redundant visibility propagation for nested elements, except where required by language specifics.
+
 ## 13.2.0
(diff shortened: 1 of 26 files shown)


Hsilgos (approved):
Didn't get why @nodoc disappeared

DanielKamkha:
Nodoc "disappeared" because it is redundant for functions and fields of a type that already has nodoc.

PR #1471, opened 2022-08-22
Add support for `@Internal` attribute

DanielKamkha:
Added support for `@Internal` attribute. Currently this attribute is
equivalent to the existing `internal` visibility modifier.


docs/lime_attributes.md
```diff
@@ -24,6 +24,8 @@ deprecated, takes a string literal value as a deprecation message. Platform-spec
 deprecation messages (see `Platform-specific comments` below for syntax).
 * **@Cached**: marks a property to be cached on platform side (i.e. read from C++ only once on first
 access and cached in Java/Swift/Dart afterwards). Currently, only supported for read-only properties.
+* **@Internal**: marks an element to have "internal" visibility. Currently, this is equivalent to the `internal`
```
Hsilgos, line 27: To which languages is it applicable? May be there is no sense in non-language attribute?

DanielKamkha: It is applicable everywhere, but C++. So a global attribute does make sense, at least to reduce verbosity. It is somewhat similar to language-based and non-language skip/enable attributes, which also mostly don't work in C++.

Hsilgos (approved):

PR #1486, opened 2022-09-05
Updated Gradle to v7.4 and AGP to v7.2.1

sameermahant:
Updated Gradle to v7.4 and AGP to v7.2.1

Signed-off-by: Sameer Mahant <3286086+sameermahant@users.noreply.github.com>

### DIFF
--- a/cmake/modules/gluecodium/gluecodium/details/gradle/wrapper/gradle-wrapper.properties
+++ b/cmake/modules/gluecodium/gluecodium/details/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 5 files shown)


cmake/modules/gluecodium/gluecodium/details/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-all.zip
```
DanielKamkha, line 3: Why "all" now instead of "bin"?

DanielKamkha (changes requested):
Also, please use your HERE email address when signing off the commit.

cmake/modules/gluecodium/gluecodium/details/gradle/wrapper/gradle-wrapper.properties
Hsilgos: We download `all` in SDK to have relevant deprecation messages and documentation. I'd say it might be also applicable for Gluecodium. But for Gluecodium it's not so important may be.

Hsilgos (commented):
LGTM, @DanielKamkha, do you still have concerns about Gradle package ?

DanielKamkha:
Yes, I'm still concerned about the size of the Gradle package. It affects the speed of the CI builds. So please use "bin", I don't have we need the Gradle docs for Gluecodium build logs.

DanielKamkha (approved):

DanielKamkha (approved):

PR #1499, opened 2022-11-24
Fixed missing setter parameter documentation in Dart

sameermahant:
Fixed missing setter parameter documentation in Dart

Signed-off-by: Sameer Mahant 3286086+sameermahant@users.noreply.github.com

### DIFF
--- a/gluecodium/src/main/resources/templates/dart/DartProperty.mustache
+++ b/gluecodium/src/main/resources/templates/dart/DartProperty.mustache
@@ -28,7 +28,8 @@
 }}{{>dart/DartFunctionBody}}
 {{/getter}}
 {{#setter}}
-{{#unless skipDocs}}{{>dart/DartDocumentation}}
+{{#unless skipDocs}}{{>dart/DartFunctionDocs}}
+
 {{/unless}}{{>dart/DartAttributes}}
 {{#if override}}{{#unless isStatic}}@override
 {{/unless}}{{/if}}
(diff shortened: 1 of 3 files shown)


gluecodium/src/main/resources/templates/dart/DartProperty.mustache
```diff
@@ -28,7 +28,8 @@
 }}{{>dart/DartFunctionBody}}
 {{/getter}}
 {{#setter}}
-{{#unless skipDocs}}{{>dart/DartDocumentation}}
+{{#unless skipDocs}}{{>dart/DartFunctionDocs}}
+
```
DanielKamkha, line 32: This blank line is extraneous, please remove it.

gluecodium/src/main/resources/templates/dart/DartProperty.mustache
```diff
@@ -28,7 +28,8 @@
 }}{{>dart/DartFunctionBody}}
 {{/getter}}
 {{#setter}}
-{{#unless skipDocs}}{{>dart/DartDocumentation}}
+{{#unless skipDocs}}{{>dart/DartFunctionDocs}}
```
DanielKamkha, line 31: Please use `{{>dart/DartFunctionDocs}}` in line 23 above as well. It does not affect the output, but it's needed for template consistency.

DanielKamkha (changes requested):

gluecodium/src/main/resources/templates/dart/DartProperty.mustache
sameermahant: This blank line is added purposely.
Without this line the output result was like:
```bash
/// [value] Some very useful property.
///set isSomeProperty(bool value) {
  ```
It was commenting the actual function definition line.

gluecodium/src/main/resources/templates/dart/DartProperty.mustache
sameermahant: If I use `DartFunctionDocs` template in `getter` code generation (line 23), It will generate duplicate documentation in getter and setter of dart language only. 
It will be inconsistent with current outputs of C++/Java/etc.

DanielKamkha (approved):

PR #1520, opened 2023-04-17
Fix for several JavaDoc warnings

LeOric32:


gluecodium/src/main/resources/templates/java/HashMapBuilder.mustache
```diff
@@ -50,6 +50,9 @@ public final class HashMapBuilder<K, V> {
 
     /**
      * @hidden
+     * @return New instance of HashMap
```
Hsilgos, line 53: @returns The reference to self for chaining.

gluecodium/src/main/resources/templates/java/HashMapBuilder.mustache
```diff
@@ -58,6 +61,7 @@ public final class HashMapBuilder<K, V> {
 
     /**
      * @hidden
+     * @return New instance of Map
```
Hsilgos, line 64: @return The built map

gluecodium/src/main/resources/templates/java/HashMapBuilder.mustache
```diff
@@ -50,6 +50,9 @@ public final class HashMapBuilder<K, V> {
 
     /**
      * @hidden
+     * @return New instance of HashMap
+     * @param key Key for object
```
Hsilgos, line 54: @param key The key with which the specified value is to be associated

Hsilgos (approved):

PR #1536, opened 2023-06-29
Add forward declarations for inner interfaces and classes in C++ code.

NazarKacharaba:


gluecodium/src/main/resources/templates/cpp/CppClass.mustache
```diff
@@ -26,6 +26,16 @@ public:
...
+{{#ifPredicate "needsInnerForwardDeclarations"}}
+public:
+{{#this.classes}}
+    class {{this.name}};
+{{/this.classes}}
+{{#this.interfaces}}
+    class {{this.name}};
+{{/this.interfaces}}
+{{/ifPredicate}}
+
```
NazarKacharaba, line 38: Here we have one problem, this will generate forward declarations for all inner classes and interfaces that are defined within a class(see [gluecodium/src/test/resources/smoke/instances/output/cpp/include/smoke/forward/InnerClassForwardDeclarations.h](https://github.com/heremaps/gluecodium/pull/1536/files#diff-80a184ea1942465c523a1ac9a2bfe2054f4ae9c84742dcfa85773d3d2583af95R21-R22)), but rather we need only those that are used in other inner classes (in our test `InnerInterface`). 

I do not have a clear solution for how to achieve this w/o using hacks like injecting this logic in the `model` or create additional `helper` for `TemplateEngine`

Hsilgos: I think we can forward declare only interface and only when classes exist. Naive implementation is currently ok.

Hsilgos (approved):

Issue #1544, opened 2023-08-04
Lambda parameters and return value docs don't work for cpp,swift and dart.


limingchina:
Currently, with the following input:
```
package test

class CommentsLambda {
    // This is a comment for lambda 
    // @param[p0] This is the comment for the first parameter of the lambda.
    // @param[p1] This is the comment for the second parameter of the lambda.
    lambda MyCallback = (@Java("param1_name") Int?, @Java("param2_name") String) 
        ->  /*Comment for return value*/ Int
} 
```
The documentation for parameters will not appear in the generated documentation for the lambda. The comment for the return value will lead to a lime file loading error since the syntax is not supporting documenting the return value.

pwrobeldev:
The support was added long time ago via https://github.com/heremaps/gluecodium/pull/1546

I am closing this ticket.

PR #1545, opened 2023-08-04
Add doc support of lambda parameters and return value for cpp,swift,dart

limingchina:
see: https://github.com/heremaps/gluecodium/issues/1544

Currently, the documentation of parameters and return values for lambdas only works for java. All the other generators don't work. This is because that Java's lambda declaration is called [functional interface](https://www.geeksforgeeks.org/functional-interfaces-java/) and is similar like a function declaration. However, for the other generators, the mustache template file only uses normal documentation instead of function documentation. 

For the return type, the issue is that the LimeParser doesn't have doc support for the return type of a lambda.

This change addresses the two issues.

The fix still has some limitation, though. In java, one can use specify a name for the lambda parameter and use that in the parameter documentation. With the current change, only the types will appear in the lambda declaration and docs. To limit the scope of this change. This is left for future improvement.

### DIFF
--- a/gluecodium/src/main/java/com/here/gluecodium/generator/cpp/CppNameResolver.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/generator/cpp/CppNameResolver.kt
@@ -85,8 +85,8 @@ internal class CppNameResolver(
             is LimeType -> resolveTypeName(element, isFullName = false)
             is LimeTypeRef -> resolveTypeRef(element)
             is LimeReturnType -> resolveTypeRef(element.typeRef)
-            is LimeLambdaParameter -> resolveTypeRef(element.typeRef)
             is LimeNamedElement -> nameCache.getName(element)
+            is LimeLambdaParameter -> resolveTypeRef(element.typeRef)
             else -> throw GluecodiumExecutionException("Unsupported element type ${element.javaClass.name}")
         }
 
--- a/gluecodium/src/main/java/com/here/gluecodium/generator/cpp/CppNameResolver.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/generator/cpp/CppNameResolver.kt
@@ -41,6 +41,7 @@ import com.here.gluecodium.model.lime.LimeExternalDescriptor.Companion.SETTER_NA
 import com.here.gluecodium.model.lime.LimeField
 import com.here.gluecodium.model.lime.LimeFunction
 import com.here.gluecodium.model.lime.LimeGenericType
+import com.here.gluecodium.model.lime.LimeLambdaParameter
 import com.here.gluecodium.model.lime.LimeList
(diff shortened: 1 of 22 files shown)


gluecodium/src/main/java/com/here/gluecodium/generator/cpp/CppNameResolver.kt
```diff
@@ -84,6 +85,7 @@ internal class CppNameResolver(
             is LimeType -> resolveTypeName(element, isFullName = false)
             is LimeTypeRef -> resolveTypeRef(element)
             is LimeReturnType -> resolveTypeRef(element.typeRef)
+            is LimeLambdaParameter -> resolveTypeRef(element.typeRef)
```
limingchina, line 88: Without this addition, it would go to the throw statement.

gluecodium/src/main/resources/templates/dart/DartLambda.mustache
```diff
@@ -18,7 +18,9 @@
   ! License-Filename: LICENSE
   !
   !}}
-{{>dart/DartDocumentation}}{{>dart/DartAttributes}}
+{{>dart/DartLambdaDocs}}
+
```
limingchina, line 22: I have to leave a blank line here. Otherwise, the lambda declaration will be in the same line as the empty comment, which would hide the lambda declaration.

gluecodium/src/main/resources/templates/dart/DartFunctionDocs.mustache
```diff
@@ -18,23 +18,8 @@
   ! License-Filename: LICENSE
   !
   !}}
-{{#resolveName comment}}{{#unless this.isEmpty}}{{prefix this "/// "}}
```
limingchina, line 21: I have to split this the doc part and attribute part. Otherwise, the lambda gets some duplicated attributes [here](https://github.com/heremaps/gluecodium/blob/46a30559cade06edd7bf56c7053c776061bda071/gluecodium/src/test/resources/smoke/attributes/output/dart/lib/src/smoke/attributes_lambda.dart#L4)

gluecodium/src/test/resources/smoke/comments/output/cpp/include/smoke/ExcludedCommentsOnly.h
```diff
@@ -28,7 +28,11 @@ class _GLUECODIUM_CPP_EXPORT ExcludedCommentsOnly {
         USELESS
     };
     /**
+     *
      * \private
+     * \param[in] ::std::string
```
limingchina, line 33: This one lost the const&. Would need some fix.

limingchina:
https://github.com/heremaps/gluecodium/pull/1546 has a more complete fix. Close this PR.

Issue #1547, opened 2023-08-18
duplicate_definition for internal classes with throwing constructors


parzysty:
For following lime file
`
package smoke

open class ExternalClass {
    enum ErrorEnum {
        NONE,
        CRASHED
    }

    class InternalOne {
        @Dart(Default)
        constructor create() throws ConstructorExploded   
    }

    class InternalTwo {
        @Dart(Default)
        constructor create() throws ConstructorExploded   
    }

    exception ConstructorExploded(ErrorEnum)
}
`

generated dart code contains multiple declaration of

```
final _$initReturnReleaseHandle = __lib.catchArgumentError(() => __lib.nativeLibrary.lookupFunction<
    Void Function(Pointer<Void>),
    void Function(Pointer<Void>)
  >('library_smoke_ExternalClass_InternalOne_create_return_release_handle'));

final _$initReturnGetResult = __lib.catchArgumentError(() => __lib.nativeLibrary.lookupFunction<
    Pointer<Void> Function(Pointer<Void>),
    Pointer<Void> Function(Pointer<Void>)
  >('library_smoke_ExternalClass_InternalOne_create_return_get_result'));

final _$initReturnGetError = __lib.catchArgumentError(() => __lib.nativeLibrary.lookupFunction<
    Uint32 Function(Pointer<Void>),
    int Function(Pointer<Void>)
  >('library_smoke_ExternalClass_InternalOne_create_return_get_error'));

final _$initReturnHasError = __lib.catchArgumentError(() => __lib.nativeLibrary.lookupFunction<
    Uint8 Function(Pointer<Void>),
    int Function(Pointer<Void>)
  >('library_smoke_ExternalClass_InternalOne_create_return_has_error'));
```

and

```
final _$initReturnReleaseHandle = __lib.catchArgumentError(() => __lib.nativeLibrary.lookupFunction<
    Void Function(Pointer<Void>),
    void Function(Pointer<Void>)
  >('library_smoke_ExternalClass_InternalTwo_create_return_release_handle'));

final _$initReturnGetResult = __lib.catchArgumentError(() => __lib.nativeLibrary.lookupFunction<
    Pointer<Void> Function(Pointer<Void>),
    Pointer<Void> Function(Pointer<Void>)
  >('library_smoke_ExternalClass_InternalTwo_create_return_get_result'));

final _$initReturnGetError = __lib.catchArgumentError(() => __lib.nativeLibrary.lookupFunction<
[...]

PR #1548, opened 2023-08-21
duplicate_definition for internal classes with throwing constructors

parzysty:
Added ffi prefix to name of functions handling error and return type retrival from throwing functions. This prevent duplicate_definition error if more than one class has those function inside one file, for example in case of multiple internal classes.

Relates-To:  #1547

### DIFF
--- a/gluecodium/src/test/resources/smoke/throwing_constructors/input/ThrowingConstructors.lime
+++ b/gluecodium/src/test/resources/smoke/throwing_constructors/input/ThrowingConstructors.lime
@@ -1,0 +1,45 @@
+# Copyright (C) 2023 HERE Europe B.V.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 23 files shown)


gluecodium/src/test/resources/smoke/throwing_constructors/input/ThrowingConstructors.lime
```diff
@@ -0,0 +1,45 @@
+# Copyright (C) 2016-2019 HERE Europe B.V.
```
sameermahant, line 1:
```suggestion
# Copyright (C) 2023 HERE Europe B.V.
```

sameermahant (approved):

NazarKacharaba:
@parzysty could you add this fix in the CHANGELOG file here - https://github.com/heremaps/gluecodium/blob/master/CHANGELOG.md?plain=1#L3 ?

sameermahant (approved):

PR #1607, opened 2024-10-25
C++: use references for non-trivial types in accessors

pwrobeldev:
The generated accessors used by-value return and parameter
types even for non-trivial types. This change is intended
to improve the situation by introducing usage of reference
for the mentioned non-trivial types.
    
This change:
 - introduces usage of const references for non-trivial types
    in getters defined via accessors attribute for structures
    when it is called on l-value
 - introduces usage of by-value type for non-trivial types
    in getters defined via accessors attribute for structures
    when it is called on r-value
 - introduces usage of const references for non-trivial types
    in setters
 - strips constness and references from type used to select a hash
    in 'StructHashImpl.mustache' because std::hash<T> does not
    accept references/constness for T
  - updates output of smoke tests to show, that since this commit
     the generated accessors use const references for non-trivial
     types

Note: the following changes extend smoke tests to show that
          accessors defined for non-trivial types use references,
          whether accessors defined for trivial types use values.


gluecodium/src/main/resources/templates/cpp/CppStruct.mustache
```diff
@@ -51,10 +51,17 @@ public:
 }}{{#if attributes.cpp.accessors}}
 {{#set structElement=this}}{{#fields}}
 {{prefixPartial "cpp/CppFieldDoc" "    "}}
-    {{resolveName typeRef}} {{resolveName this "" "getter"}}( ) const { return {{resolveName}}; }{{!!
+    {{#ifPredicate typeRef "needsRefSuffix"}}const {{resolveName typeRef}}&{{!!
+}} {{resolveName this "" "getter"}}( ) const & { return {{resolveName}}; }
+    {{resolveName typeRef}} {{resolveName this "" "getter"}}( ) const && { return {{resolveName}}; }{{/ifPredicate}}{{!!
```
Hsilgos, line 56:
```suggestion
    const {{resolveName typeRef}}&&{{resolveName this "" "getter"}}( ) const && { return std::move({{resolveName}}); }{{/ifPredicate}}{{!!
```

pwrobeldev: Fixed the problem by introducing 3 overloads of getter for non-trivial types.

Moreover, for such types 2 overload of setters have been added -- `const lvalue` ref and `rvalue ref`.

pwrobeldev:
Note for reviewers -- the change is split into 3 commits:
- extension of smoke tests to show invalid behavior of generator of getters -- it is big, but contains mostly the generated code
- the actual fix for the problem -- it is not so big and that's the core change in logic
- implementation of a new functional test -- for further verification

gluecodium/src/main/resources/templates/cpp/CppStruct.mustache
pwrobeldev: The code was updated according to the offline discussion.

functional-tests/functional/cpp/tests/AccessorsTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+    auto list_field = get_const_struct().get_nontrivial_list_field();
+
+    // Then access can be made.
+    EXPECT_EQ(TrivialInt{64}, int_field);
+    EXPECT_DOUBLE_EQ(TrivialDouble{52.3}, double_field);
+    EXPECT_EQ(NontrivialString{"Another string"}, string_field);
+
+    const NontrivialList expected_list{"A1", "A2", "A3"};
+    EXPECT_EQ(expected_list, list_field);
+}
```
pwrobeldev, line 135: Note: if these test cases used `auto&&`, then we would have dangling references.

Hsilgos (approved):

PR #1608, opened 2024-10-30
C++: add missing export macro to 'CppEnumeration.mustache'

pwrobeldev:
Generated files for C++ did not contain export macro for enumerations.
That was not the case of structs and classes. The reason for that was
missing inclusion of 'CppExportMacro' in 'CppEnumeration.mustache'.

Note: the change in logic is only a single line.
The remaining changes in the commit are related
to smoke tests -- the expected generated files
with enumerations have been adjusted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Bug fixes:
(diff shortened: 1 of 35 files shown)


pwrobeldev:
It turned out, that the change is not needed.

PR #1553, opened 2023-09-15
Clean up the generated code for java.

NazarKacharaba:
Clean up the generated code for Java: remove blank lines between javadoc and method signature.

PR #1610, opened 2024-10-31
Dart: generate omitted documentation of properties

pwrobeldev:
When users want to document a property
in LIME files, they can use structured
comments, which consist of the general
information about property and annotations
related to getters and setters.

In the case of Dart generator, the output
contained only comments related to getters
and setters. The general information about
property was not stored in the output files.

This change adjusts 'DartRedirectProperty'
mustache template to output also the general
information about property if present.

The behaviour in the case of Dart generator
has been aligned with C++ and Java generators.
The general information about property is added
to comments of getter and setter.

### DIFF
--- a/gluecodium/src/test/resources/smoke/attributes/output/dart/lib/src/smoke/attributes_with_comments.dart
+++ b/gluecodium/src/test/resources/smoke/attributes/output/dart/lib/src/smoke/attributes_with_comments.dart
@@ -1,34 +1,48 @@
+
+
 import 'dart:ffi';
 import 'package:library/src/_library_context.dart' as __lib;
 import 'package:library/src/_native_base.dart' as __lib;
 import 'package:library/src/_token_cache.dart' as __lib;
 import 'package:library/src/builtin_types__conversion.dart';
+
 /// Class comment
 @OnClass
 abstract class AttributesWithComments {
+
   /// Const comment
   @OnConstInClass
   static final bool pi = false;
+
   /// Function comment
   ///
   @OnFunctionInClass
   void veryFun();
+  /// Property comment
   /// Getter comment
   @OnPropertyInClass
   String get prop;
+  /// Property comment
   /// Setter comment
   @OnPropertyInClass
   set prop(String value);
+
 }
+
+
 class AttributesWithComments_SomeStruct {
   /// Field comment
   @OnField
   String field;
+
   AttributesWithComments_SomeStruct._(this.field);
   AttributesWithComments_SomeStruct()
     : field = "";
 }
+
+
 // AttributesWithComments_SomeStruct "private" section, not exported.
+
 final _smokeAttributeswithcommentsSomestructCreateHandle = __lib.catchArgumentError(() => __lib.nativeLibrary.lookupFunction<
     Pointer<Void> Function(Pointer<Void>),
     Pointer<Void> Function(Pointer<Void>)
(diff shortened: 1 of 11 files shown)


kcherneha:
Across almost all files I see redundant new lines, could you please remove them?
As well as {{escape ""}} - not sure if needed, 

```

  /// Property comment
  /// Description of: String prop.
  --- here is empty line - I don't think we need it?
  /// Getter comment
  @OnPropertyInClass
```

pwrobeldev:
> Across almost all files I see redundant new lines, could you please remove them? As well as {{escape ""}} - not sure if needed,
> 
> ```
> 
>   /// Property comment
>   /// Description of: String prop.
>   --- here is empty line - I don't think we need it
>   /// Getter comment
>   @OnPropertyInClass
> ```

Fixed.

My initial idea was to separate the field description from the comment that describes getter. Because after the proposed change the general part of property comment is a part of getter description in Dart. However, if this is the requirement, then I am okay with it.

In C++/Java both getter and setter have the part related to the general comment about property. If needed, this can be also added to setter in Dart -- the change will be 1 line. 

Because right now the behavior is as follows:
- getter in Dart contains `@get` comment as well as the general comment
- setter in Dart contains only `@set` comment

pwrobeldev:
Also, regarding the empty lines -- Smoke tests do not require updating files when only whitespaces change. Therefore, my change accumulated also changes in whitespaces that were done long time ago. 

Please check: `private fun ignoreUnimportantDifferences(text: String)` function that is present in SmokeTest.kt.

kcherneha:
> > Across almost all files I see redundant new lines, could you please remove them? As well as {{escape ""}} - not sure if needed,
> > ```
> > 
> >   /// Property comment
> >   /// Description of: String prop.
> >   --- here is empty line - I don't think we need it
> >   /// Getter comment
> >   @OnPropertyInClass
> > ```
> 
> Fixed.
> 
> My initial idea was to separate the field description from the comment that describes getter. Because after the proposed change the general part of property comment is a part of getter description in Dart. However, if this is the requirement, then I am okay with it.
> 
> In C++/Java both getter and setter have the part related to the general comment about property. If needed, this can be also added to setter in Dart -- the change will be 1 line.
> 
> Because right now the behavior is as follows:
> 
> * getter in Dart contains `@get` comment as well as the general comment
> * setter in Dart contains only `@set` comment

Ideally, Swift, Java and Dart (from comments perspective) should be aligned as much as possible. If it doesn't bring redundancy and good for readability - you can add it as well.

gluecodium/src/test/resources/smoke/attributes/output/dart/lib/src/smoke/attributes_with_comments.dart
Hsilgos, line 1: Is here any changes?

Hsilgos (dismissed):

gluecodium/src/test/resources/smoke/attributes/output/dart/lib/src/smoke/attributes_with_comments.dart
pwrobeldev: Only line 21: `/// Property comment`

kcherneha (approved):

Hsilgos:
> Across almost all files I see redundant new lines, could you please remove them? As well as {{escape ""}} - not sure if needed,
> 
> ```
> 
>   /// Property comment
>   /// Description of: String prop.
>   --- here is empty line - I don't think we need it
>   /// Getter comment
>   @OnPropertyInClass
> ```

Empty lines were introduced not by this change. Some empty lines should be removed with https://github.com/heremaps/gluecodium/pull/1553 but generally speaking either we have empty lines or more clean mustache files.

Hsilgos (approved):

PR #1611, opened 2024-11-04
Docs: document Dart-specific behavior of internal tag

pwrobeldev:
The behavior of internal tag for Dart is different from that for Java/Swift. 
In Dart, the types annotated as internal do have internal visibility.

However, functions, properties, constructors and field constructors are
not generated. These elements are skipped like in the case of skip attribute.

This change adds this information, because the discrepancy between languages
may be surprising for the users.


pwrobeldev:
Just to have links to the issue and the functional change, that introduced this behavior:
- GitHub issue related to `@Internal` tag in Dart: https://github.com/heremaps/gluecodium/issues/1333
- PR that introduced skipping generation of internal functions, properties, constructors and field constructors: https://github.com/heremaps/gluecodium/pull/1379

Hsilgos (approved):

PR #1612, opened 2024-11-05
Fix conversion of platforms for 'Internal' annotation

pwrobeldev:
The conversion of platforms for 'Internal' annotation was invalid
when they were specified without quotation marks. In such a case
the converter set internal visibility for all platforms.
    
This change adjusts the converter as well as the output of smoke
tests to show that the behavior is correct now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased:
 ### Bug fixes:
+ * Fixed conversion of platforms for 'Internal' annotation to work correctly also when arguments are not in quotation marks.
  * Dart: generate omitted documentation of properties.
 
 ## 13.9.6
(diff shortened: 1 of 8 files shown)


pwrobeldev:
For the first commit, which shows the invalid behavior please see the following input:

```
@Internal(Java)
class JavaInternalClassRev {}

@Internal(Swift)
class SwiftInternalClassRev {}

@Internal(Dart)
class DartInternalClassRev {}

@Internal(Java, Swift)
class JavaSwiftInternalClass {}
```

Even though platforms are specified, the behavior without fix was incorrect -- the classes were internal for all platforms.

With the fix they are internal only for the specified platforms.

Hsilgos (approved):

PR #1614, opened 2024-11-06
Ensure that common utilities are always exported

pwrobeldev:
If the user builds the code with visibility hidden,
then certain common utilities may not be visible.
    
This change adds appropriate export macro to the
required functions and types to ensure, that they
are always exported.

### DIFF
--- a/gluecodium/src/main/resources/templates/ffi/FfiDartDLInitImpl.mustache
+++ b/gluecodium/src/main/resources/templates/ffi/FfiDartDLInitImpl.mustache
@@ -1,6 +1,6 @@
 {{!!
   !
-  ! Copyright (C) 2016-2021 HERE Europe B.V.
+  ! Copyright (C) 2016-2024 HERE Europe B.V.
   !
   ! Licensed under the Apache License, Version 2.0 (the "License");
   ! you may not use this file except in compliance with the License.
@@ -19,6 +19,8 @@
   !
   !}}
 {{>ffi/FfiCopyrightHeader}}
+
+#include "Export.h"
 
 #include "DartDLInit.h"
 #include "dart_api_dl.h"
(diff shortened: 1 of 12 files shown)


gluecodium/src/main/resources/templates/ffi/FfiDartDLInitImpl.mustache
```diff
@@ -20,6 +20,9 @@
   !}}
 {{>ffi/FfiCopyrightHeader}}
 
+// Ensure, that symbols for function pointers are externally visible.
+#define DART_SHARED_LIB
```
pwrobeldev, line 24: It shall not be defined here.

The user could supply it via CMake definition.

gluecodium/src/main/resources/templates/jni/utils/JniBaseHeader.mustache
```diff
@@ -28,7 +28,7 @@
 // JNI_OnLoad to allow for a custom implementation or similar. In that case the user
 // needs to ensure it is called manually. If it has a custom name, declare the function.
 #ifdef GLUECODIUM_JNI_ONLOAD
-jint GLUECODIUM_JNI_ONLOAD( JavaVM* vm, void* );
+JNIEXPORT jint GLUECODIUM_JNI_ONLOAD( JavaVM* vm, void* );
 #else
 # define GLUECODIUM_JNI_ONLOAD JNI_OnLoad
```
pwrobeldev, line 33: Perhaps, the export is also needed when non-custom version of the function is defined.

pwrobeldev:
As expected: adding `-no-undefined` flag to linker uncovered missing symbols related to JNI and Dart common code. In next patches fixes will be provided.

pwrobeldev:
> As expected: adding `-no-undefined` flag to linker uncovered missing symbols related to JNI and Dart common code. In next patches fixes will be provided.

With next patches the errors were resolved.

Hsilgos (approved):

PR #1617, opened 2024-11-14
Improve structured comments syntax for properties 

pwrobeldev:
The first part of comment used to describe a property (lines before get/set annotations)
was used in a few places:
- to document the property when declaring it in Swift and Dart
- to document the parameter used in setter in the case of C++ and Java
- to document the return value of getter in the case of C++ and Java

In some cases it looked strange. For instance, when the user wanted to leave some additional,
extended information about the property, that was necessary to use it correctly, then it was
appended to the description of the return type and parameters of getters and setters.

Moreover, it was hard for users to remember how these lines are rendered in the case of four
output languages.

In order to ease the documentation of properties, this change introduces two new annotations:
'value' and 'description'. The structured comments for properties can be formed as follows:

`@value`:
- **purpose:** short description of property (ideally single line)
- **usage:** declaration of property (Swift/Dart) or setter_param/getter_return (Java/C++)

`@description`:
- **purpose:** extended information about property that is required to use it correctly
- **usage:** declaration of property (Swift/Dart), declaration of getter/setter (Java/C++)

`@get`:
- **purpose:** documents getter function
- **usage:** declaration of getter function

`@set`:
- **purpose:** documents setter function
- **usage:** declaration of setter function

Moreover, the validation was implemented. The rules are as follows:
- the user cannot start a comment without annotation and then use later `@value`
- if the user forgets to use `@value` then warning will be generated
- if the user wants, the warning related to missing `@value` can be treated as error via `werror` flag
- the old way of generating comment will be used when the comment does not use `@value` but provides
  information without annotation -- however, the warning/error will be generated as well

### DIFF
--- a/gluecodium/src/main/java/com/here/gluecodium/validator/LimePropertiesValidator.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/validator/LimePropertiesValidator.kt
@@ -88,8 +88,8 @@ internal class LimePropertiesValidator(private val logger: LimeLogger, generator
         val checkDocsMessage = "please check $PROPERTIES_STRUCTURED_COMMENTS"
         var result = true
 
-        if (limeProperty.comment.isEmpty()) {
-            logger.maybeError(limeProperty, "property description comment is missing; $checkDocsMessage")
+        if (limeProperty.valueComment.isEmpty()) {
+            logger.maybeError(limeProperty, "property must use @value annotation; $checkDocsMessage")
             result = false
         }
 
--- a/gluecodium/src/test/java/com/here/gluecodium/validator/LimePropertiesValidatorTest.kt
+++ b/gluecodium/src/test/java/com/here/gluecodium/validator/LimePropertiesValidatorTest.kt
@@ -76,7 +76,7 @@ class LimePropertiesValidatorTest {
         typeRef = LimeBasicTypeRef.INT,
         getter = getter,
         setter = setter,
-        comment = valueComment,
+        valueComment = valueComment,
     )
 
     private fun getterForPath(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased:
 ### Features:
+ * Implemented the new way of documenting properties via `@value` and `@description` annotations in structured comments.
+ * Implemented validation, which requires the users to use at least `@value` annotation when documenting a property.
  * Implemented validation of comments used for accessors of properties: parameter of setter and return value of getter. When the required description is missing, then warning is generated. The user may also treat the warning as error via 'werror' flag.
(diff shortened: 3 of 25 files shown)


gluecodium/src/main/java/com/here/gluecodium/validator/LimePropertiesValidator.kt
```diff
@@ -63,6 +67,12 @@ internal class LimePropertiesValidator(private val logger: LimeLogger) {
             logger.error(limeProperty, "property in an interface cannot be marked with @Cpp(Ref)")
             result = false
         }
+        if (limeProperty.valueComment.isEmpty()) {
```
pwrobeldev, line 70: I wonder whether this is an optimal solution.

If the user does not add any comment to the property, then this still prints a warning.


In the beginning I considered to print the warning/error only when user adds a comment, but it does not use `@value` annotation. It can be easily achieved:
`if (limeProperty.valueComment.isEmpty() && !limeProperty.comment.isEmpty())`

What do you think about it? Do many users want to skip adding any documentation to property altogether?

pwrobeldev:
Note for reviewers: 
- it is much easier to review this change when going commit-by-commit to see why certain changes were made

gluecodium/src/main/java/com/here/gluecodium/validator/LimePropertiesValidator.kt
Hsilgos: I think this warning make sense for a public property. I.e. if a property is internal (or class/struct which has the property is internal, etc) then it's ok to skip documentation.

pwrobeldev: Implemented the logic according to the comment. Added also unit tests that check the mentioned path.

gluecodium/src/test/java/com/here/gluecodium/validator/LimePropertiesValidatorTest.kt
pwrobeldev, line 1: Note: I had to add "more realistic" paths to elements in test cases, because without it the code could fail when checking if property needs the documentation comment. 


`AntlrLimeModelBuilder` uses `SomeType.SomeProperty` as property path. 

```
    override fun enterProperty(ctx: LimeParser.PropertyContext) {
        pushPath(ctx.simpleId())
        structuredCommentsStack.push(parseStructuredComment(ctx.docComment(), ctx))
    }
```

and then:

```
    private fun pushPath(simpleId: LimeParser.SimpleIdContext) {
        val newPath = currentPath.child(convertSimpleId(simpleId))
        pathStack.push(referenceResolver.computeUniquePath(newPath))
    }
```

Hsilgos (approved):

PR #1618, opened 2024-11-18
Warn when the required documentation for functions is missing

pwrobeldev:
This change extends function validator by checking
if parameters and return value are properly documented.
    
The logic works as follows:
 - all function parameters need to be documented
 - return value needs to be documented
 - the validation is not applied for internal functions
   or functions nested in internal types (or types nested
   in internal types)
    
The user may enable treating the warning as error via werror
flag.

### DIFF
--- a/gluecodium/src/main/java/com/here/gluecodium/Gluecodium.kt
+++ b/gluecodium/src/main/java/com/here/gluecodium/Gluecodium.kt
@@ -193,7 +193,7 @@ class Gluecodium(
             { LimeConstantRefsValidator(limeLogger).validate(it) },
             { LimeExternalTypesValidator(limeLogger).validate(it) },
             { LimePropertiesValidator(limeLogger).validate(it) },
-            { LimeFunctionsValidator(limeLogger).validate(it) },
+            { LimeFunctionsValidator(limeLogger, generatorOptions).validate(it) },
             { LimeSkipValidator(limeLogger).validate(it) },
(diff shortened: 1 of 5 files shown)


pwrobeldev:
Please see how the output looks like. For instance, in smoke tests we have `ForwardDeclarations.lime` file, which contains the following interface:

```
interface UseForward {
    fun use_it(
        param1: Class1,
        param2: Class2,
        simpleClass: smoke.SimpleClass,
        simpleInterface: smoke.SimpleInterface
    )
}
```

The documentation is missing for parameters and return value.


When the code is generated, the following warning is produced.

```
ForwardDeclarations.lime, element smoke.forward.UseForward.use_it: Parameter 'param1' must be documented
ForwardDeclarations.lime, element smoke.forward.UseForward.use_it: Parameter 'param2' must be documented
ForwardDeclarations.lime, element smoke.forward.UseForward.use_it: Parameter 'simpleClass' must be documented
ForwardDeclarations.lime, element smoke.forward.UseForward.use_it: Parameter 'simpleInterface' must be documented
ForwardDeclarations.lime, element smoke.forward.UseForward.use_it: Return must be documented
```

Hsilgos (dismissed):

Hsilgos (approved):

pwrobeldev:
`CHANGELOG.md` was not adjusted. It is done by the follow-up PR: https://github.com/heremaps/gluecodium/pull/1620

PR #1622, opened 2024-11-21
Dart: fix problems with default values that use collection types

pwrobeldev:
This series fixes problems that occur when either `@Immutable`
or `@PositionalDefaults` annotation is used in combination with
collection types.

-------- `@PositionalDefaults` --------
The usage of 'PositionalDefaults' annotation
combined with nullable collection field that has 'null'
default value yields the code that does not compile.

The generator blindly applies 'const' keyword before
the default value for collections. A special 'null' value is
not taken into account.

-------- `@Immutable` --------
Usage of 'Immutable' annotation combined
with collection fields that have default values yields the
code that does not compile.
    
The generator does not apply 'const' keyword before the calls
to constructors of collections. It is required, because the
constructor of 'Immutable' structure is marked as 'const'.
    
The same happens when an explicit field constructor is used.
Because of 'Immutable' usage it is marked as a const, but
the default values are not initialized with values prepended
by 'const'.

-------- Content of change --------
- New smoke and functional tests that were used to confirm
   the invalid behavior.
- Adjustments of 'DartStructConstructors.mustache' file to
   correctly prepend 'const' keyword when needed and to treat
   'null' as it should be treated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unreleased:
 ### Features:
  * Implemented validation of comments used for functions. When the description of parameters or return value is missing, then warning is generated. The user may also treat the warning as error via 'werror' flag.
+### Bug fixes:
+ * Dart: fixed a bug related to missing/superflous 'const' keyword usage in definition of default values in constructors that used collections when `@Immutable` or `PositionalDefaults` were specified.
 
 ## 13.9.7
 Release date 2024-11-13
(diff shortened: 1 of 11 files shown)


pwrobeldev:
Note for reviewers:
- two first commits were used to cause functional tests build failure that confirmed invalid behavior of Dart generator
- then the adjustments were implemented in mustache templates to ensure that the tests compile without problems

PR #1625, opened 2024-11-25
Gluecodium.kt: validate functions just once

pwrobeldev:
'LimeFunctionsValidator' uses 'typeRef' member while performing validation.
Therefore, it shall be present only in 'getTypeRefDependentValidators()'.

However, it is present also in 'getIndependentValidators()' which means
that Gluecodium validates functions twice.

The investigation of GIT history revealed that addition of functions validator
to 'independent validators' list was done accidentally.

This change ensures, that the mentioned validator is run only once.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,12 @@
 
 ## Unreleased:
 ### Features:
+ * Implemented validation of comments used for lambdas. When the description of parameters or return value is missing, then warning is generated. The user may also treat the warning as error via 'werror' flag.
  * Implemented validation of comments used for functions. When the description of parameters or return value is missing, then warning is generated. The user may also treat the warning as error via 'werror' flag.
 ### Bug fixes:
  * Gluecodium.kt: fixed a bug related to running 'LimeFunctionsValidator' twice. It was not needed - running the validator once is sufficient.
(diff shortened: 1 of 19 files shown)


pwrobeldev:
Hi, please find the explanation and links below.

There are two functions that return validators:
 - `getTypeRefDependentValidators()`
 - `getIndependentValidators()`

As the name suggests, the former ones are run conditionally — only when `LimeTypeRefsValidator` yields success. 
The latter ones are always run.


What brought my attention is the fact, that `LimeFunctionsValidator` is present in both lists.
That seems a little bit strange, because it means that we perform the same validation for functions twice.
Moreover, I can see that `LimeFunctionsValidator.validateFunction()` uses `typeRef` member. So, it does not fit to independent validators.


I looked into git history and it seems that addition of this validator to `independent validators` list was a mistake. The author of commit wanted to add `LimeFieldsValidator`. The commit can be found here: https://github.com/heremaps/gluecodium/commit/4e026cb22a0c973c95ed5849399b0a2067af7039

Later, the correct validator was added in bug-fix commit (link below), but addition of `LimeFunctionsValidator` to `independent validators` was not removed: https://github.com/heremaps/gluecodium/commit/972c588cb96ba2cc626d175b3f5ea35ec98e1eb1

Hsilgos (dismissed):

Hsilgos (approved):

PR #1626, opened 2024-11-25
Warn when documentation of lambda parameters is missing

pwrobeldev:
This change adds a new class called LimeLambdaValidator
which checks if parameters and return value are properly
documented.
    
The logic works as follows:
  - all lambda parameters need to be documented
  - return value needs to be documented (except 'void')
  - the validation is not applied for internal lambdas
     or lambdas nested in internal types (or types nested
     in internal types)
    
The user may enable treating the warning as error via
werror flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased:
 ### Features:
+ * Implemented validation of comments used for lambdas. When the description of parameters or return value is missing, then warning is generated. The user may also treat the warning as error via 'werror' flag.
  * Implemented validation of comments used for functions. When the description of parameters or return value is missing, then warning is generated. The user may also treat the warning as error via 'werror' flag.
 ### Bug fixes:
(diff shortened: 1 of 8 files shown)


pwrobeldev:
Please see the example output below.

In smoke tests in Lambdas.lime we have the following code:

```
lambda Producer = () -> String

[...]

lambda Consumer = (String) -> Void
```

The warnings look as follows:

```
Lambdas.lime, element smoke.Lambdas.Producer: Return must be documented

Lambdas.lime, element smoke.Lambdas.Consumer: Parameter 'p0' must be documented
```


Additionally, in CommentsLambda.lime we have the following lambda with named params that misses docs:
```
lambda NoCommentsWithNamedParams = (first: String, second: String) -> String
```

The warning is as follows:

```
CommentsLambda.lime, element smoke.LambdaComments.NoCommentsWithNamedParams: Parameter 'first' must be documented
CommentsLambda.lime, element smoke.LambdaComments.NoCommentsWithNamedParams: Parameter 'second' must be documented
CommentsLambda.lime, element smoke.LambdaComments.NoCommentsWithNamedParams: Return must be documented
```

Hsilgos (approved):

PR #1633, opened 2024-12-11
Dart: fix race condition between garbage collection of finalizable objects and native function calls

pwrobeldev:
----- Motivation -----
If the generated class is not marked with 'Finalizable' interface,
then, when methods are invoked on its object, the object can be
garbage collected in the middle of method execution.
    
For instance let's take a look at the following example from functional
tests:
    
'AsyncClass().asyncVoid(false)'
    
1. AsyncClass object is created and its lifetime is tied with the native
    finalizer, which deletes shared_ptr (this.handle) that owns resources.
2. Then, inside 'asyncVoid(false)' we use only 'this.handle' member to
    pass the opaque handle to the native function. For garbage collector
    after the handle is passed 'this' object is no longer used by anybody.
    Therefore, it is not needed – can be finalized.
3. If the garbage collection kicks in before the native call, then the
    'AsyncClass' object is garbage collected, its finalizer is run and therefore,
    the resources on C++ side are deleted. It causes the segmentation fault.
    
We have a race condition between the garbage collection and execution
of native functions.
    
The consequences can be dramatic -- the finalizer registered for
the object may be executed before the native function call inside
the member methods. The finalizers delete C++ resources tied with
the object. This can lead to segmentation fault.

----- Solution -----
The 'Finalizable' interface is a viable solution proposed by Dart:FFI
documentation. It ensures, that 'this' object outlives the whole method
call -- it guarantees, that the native function call finishes before the
finalizer is run.

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -257,7 +257,7 @@ jobs:
       - name: Install Dart SDK
         run: |
           DART_RELEASE_CHANNEL=stable
-          DART_VERSION=2.12.0
+          DART_VERSION=2.18.1
           wget -nv https://storage.googleapis.com/dart-archive/channels/${DART_RELEASE_CHANNEL}/release/${DART_VERSION}/linux_packages/dart_${DART_VERSION}-1_amd64.deb
           sudo apt -y install ./dart_${DART_VERSION}-1_amd64.deb
       - name: Build and run functional tests
@@ -373,7 +373,7 @@ jobs:
       - name: Install Dart SDK
         run: |
           DART_RELEASE_CHANNEL=stable
-          DART_VERSION=2.12.0
+          DART_VERSION=2.18.1
           wget -nv https://storage.googleapis.com/dart-archive/channels/${DART_RELEASE_CHANNEL}/release/${DART_VERSION}/linux_packages/dart_${DART_VERSION}-1_amd64.deb
           sudo apt -y install ./dart_${DART_VERSION}-1_amd64.deb
       - name: Build and run functional tests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Gluecodium project Release Notes
 
+## Unreleased
+### Bug fixes:
+ * Dart: fixed race condition between garbage collector (which could trigger finalizer) and native method calls in finalizable classes.
+
 ## 13.10.0
 Release date 2024-11-28
 ### Features:
--- a/gluecodium/src/main/resources/templates/dart/DartClass.mustache
+++ b/gluecodium/src/main/resources/templates/dart/DartClass.mustache
@@ -1,6 +1,6 @@
 {{!!
   !
-  ! Copyright (C) 2016-2019 HERE Europe B.V.
+  ! Copyright (C) 2016-2024 HERE Europe B.V.
   !
   ! Licensed under the Apache License, Version 2.0 (the "License");
   ! you may not use this file except in compliance with the License.
@@ -20,7 +20,7 @@
   !}}
 {{>dart/DartDocumentation}}{{>dart/DartAttributes}}
 abstract class {{resolveName}}{{!!
-}}{{#if this.parents}} implements {{#this.parents}}{{resolveName}}{{#if iter.hasNext}}, {{/if}}{{/this.parents}}{{/if}} {
+}} implements {{#if this.parents}}{{#this.parents}}{{resolveName}}, {{/this.parents}}{{/if}}Finalizable {
 {{#set parent=this container=this}}{{#constructors}}
 {{prefixPartial "dart/DartFunctionDocs" "  "}}
   factory {{resolveName parent}}{{>dart/DartConstructorName}}({{!!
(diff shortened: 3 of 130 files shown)


pwrobeldev:
The problem was reproducible in functional tests. On average 1 in ~15 runs failed.

Please find a few stack traces from functional tests, which show the invalid behavior.

Async_test.dart:
```
5   libfunctional.dylib           	       0x1054a9ea0 std::__1::__shared_count::__add_shared[abi:ue170006]() + 28
6   libfunctional.dylib           	       0x1054a9e78 std::__1::__shared_weak_count::__add_shared[abi:ue170006]() + 24
7   libfunctional.dylib           	       0x1059ef35c std::__1::shared_ptr<test::AsyncClass>::shared_ptr[abi:ue170006](std::__1::shared_ptr<test::AsyncClass> const&) + 92
8   libfunctional.dylib           	       0x105625b38 std::__1::shared_ptr<test::AsyncClass>::shared_ptr[abi:ue170006](std::__1::shared_ptr<test::AsyncClass> const&) + 36
9   libfunctional.dylib           	       0x105624cd0 lorem_ipsum::test::ffi::Conversion<std::__1::shared_ptr<test::AsyncClass>, void>::toCpp(void*) + 56
10  libfunctional.dylib           	       0x105624ba8 functional_test_AsyncClass_asyncVoid__asyncVoid__resultLambda_Boolean + 72
```

Inheritance_test.dart:
```
5   libfunctional.dylib           	       0x107227c64 std::__1::__shared_count::__add_shared[abi:ue170006]() + 28
6   libfunctional.dylib           	       0x107227c3c std::__1::__shared_weak_count::__add_shared[abi:ue170006]() + 24
7   libfunctional.dylib           	       0x10722d99c std::__1::shared_ptr<test::SimpleInstantiableTwo>::shared_ptr[abi:ue170006](std::__1::shared_ptr<test::SimpleInstantiableTwo> const&) + 92
8   libfunctional.dylib           	       0x10722b0e0 std::__1::shared_ptr<test::SimpleInstantiableTwo>::shared_ptr[abi:ue170006](std::__1::shared_ptr<test::SimpleInstantiableTwo> const&) + 36
9   libfunctional.dylib           	       0x107445de0 functional_test_SimpleInstantiableTwo_copy_handle + 108
```

ExternalTypes_test.dart:
```
5   libfunctional.dylib           	       0x106774670 std::__1::__shared_count::__add_shared[abi:ue170006]() + 28
6   libfunctional.dylib           	       0x106774648 std::__1::__shared_weak_count::__add_shared[abi:ue170006]() + 24
7   libfunctional.dylib           	       0x106dc5a58 std::__1::shared_ptr<test::UseMyClass>::shared_ptr[abi:ue170006](std::__1::shared_ptr<test::UseMyClass> const&) + 92
8   libfunctional.dylib           	       0x1069bfecc std::__1::shared_ptr<test::UseMyClass>::shared_ptr[abi:ue170006](std::__1::shared_ptr<test::UseMyClass> const&) + 36
9   libfunctional.dylib           	       0x1069bfc84 lorem_ipsum::test::ffi::Conversion<std::__1::shared_ptr<test::UseMyClass>, void>::toCpp(void*) + 56
10  libfunctional.dylib           	       0x1069bfb78 functional_test_UseMyClass_callBar__MyClass + 48
```


In all mentioned cases accessing the converted opaque handle caused segmentation faults. The memory was freed because of the race condition.

Please find some debug logs from one of failed cases:
```
1: 00:00 +0: Async: Async void
1: 
1: [DART] create(): handle: = 0x600003c18250
1: [DART] _asyncVoid__async(): _handle: = 0x600003c18250
1: 
1: 
1: [C++] Finalizing handle: 0x600003c18250
1: [C++] call function -- handle: 0x600003c18250
1: 
1: ===== CRASH =====
1: si_signo=Segmentation fault: 11(11), si_code=SEGV_ACCERR(2), si_addr=0x803
```

The above is an evidence of invalid behavior.

pwrobeldev:
In order to fix the problem the `Finalizable` marker interface can be used. We are hitting the exact problem, which is solved by `Finalizable` and described in the documentation. 

According to the documentation:

>Methods on a class implementing Finalizable keep the this object alive for the duration of the method execution. The this value is treated like a local variable.
>
>For example, this is kept alive during the execution of someNativeCall in myFunction:
```
class MyFinalizable implements Finalizable {
  final Pointer nativeResource;

  MyFinalizable(this.nativeResource);

  void myFunction() {
    someNativeCall(nativeResource);
  }
}

void someNativeCall(Pointer nativeResource) {
  // ..
}
```

Reference:
- https://api.flutter.dev/flutter/dart-ffi/Finalizable-class.html

pwrobeldev:
A few notes for reviewers:

- the functional tests were run 100+ times with the fix and they passed each and every time; without the fix the failure was visible once in ~15 runs -- therefore, I assume that the race is fixed
- Dart SDK version used on CI was upgraded from 2.12 to 2.18, because `Finalizable` has been available since 2.17 -- it is a minor upgrade -- the last stable version is 3+
- The actual changes in mustache templates are tiny -- the remaining changes are smoke tests, which are generated


The PR is intended to be reviewed commit-by-commit. The non-interesting changes related to the output of smoke tests are present in separate commits.

diachenko-mischa (approved):

PR #1641, opened 2024-12-18
Dart: do not generate `default` clause for enumerations

pwrobeldev:
----- Motivation -----
In the case of enumerations we match all the possible
matchable values in `case` clauses. Then, we generate
`default` clause, but the linter is able to see, that
this is a dead code.
    
In such case `unreachable_switch_default` warning is
raised by the linter.
    
----- Important -----
The above is true if we generate an authentic `enum` in Dart.
In the case when aliasing enumerators are needed, then
we generate an enum-like class. In such case `default` clause
is needed, because we operate on integers and compiler assumes,
that switch-case is not exhaustive.
    
----- Implemented solution -----
Removed generation of redundant `default` clause when we use
an authentic Dart `enum` to avoid linter warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Gluecodium project Release Notes
 
+## Unreleased
+### Bug fixes:
+ * Dart: removed generation of redundant `default` clause in exhaustive switch-cases for enumerations. The redundant `default` caused linter warnings.
+
 ## 13.10.1
 Release date 2024-12-12
 ### Bug fixes:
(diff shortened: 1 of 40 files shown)


pwrobeldev:
Dart SDK issue and code, which introduced the check for `default` clause. The issue reported the missing warning and introduced it. Links can be found here:
- https://github.com/dart-lang/sdk/issues/54575
- https://github.com/dart-lang/sdk/commit/cda2815bb1f028b4d88f664670865968bc64f847

Hsilgos (approved):

Issue #1028, opened 2021-08-05
Generate Dart "testable" code for Flutter testability


Hsilgos:
# Flutter testing

Flutter unit tests are usually ran on host. Also it's important to disconnect business logic from generated interface to avoid side effects. Following options were investigated.

## Dependency injection
There are few external dart libraries which allow to inject dependencies, for example [get_it](https://pub.dev/packages/get_it)
Unfortunately they require to write or adopt existing code in specific way.

There is no way to replace (inject) package in tests with other package (for example with stubs) like it can be done for Java.

## Conditional import

Dart 1.x supported conditional import based on specified compilation flags.
Dart 2.x is not so flexible and supports only limited conditional import based on presence of platform libraries `dart.library.xxx`, like
```
import "src/default.dart" 
  if (dart.library.io) "src/io.dart"
  if (dart.library.html) "src/web.dart"
  as impl;
```
I haven't found a way to use this approach to make conditional imports only for unit tests.

## Override lookupFunction

Since all FFI functions are resolved with `nativeLibrary.lookupFunction`, this `lookupFunction` (or wrapper around this function) can be overridden and return mocked functions. Such approach could allow to not touch existing code, but provide generated code as extra plugin for testing which replaces `__lib.nativeLibrary` with overridden version. 

The problem here is to return test data from such mocked functions. The functions work with data which is prepared for ffi call, for example `Pointer<Void>` instead of usual types. This `Pointer<Void>` must be converted back to the instance of type, perhaps with call to existing generated converter function.

It looks very costly to generate the code which allows to mock functions in convenient way.

TBD

## Factory methods

The common approach is to call factory method to select real implementation or mocking runtime. In unit tests those factory methods must be replaced during fixture setup. 

```
abstract class Foo {
    factory Foo(A a, B b, C c) => _factoryMake()

    static void doSomething() => _factoryDoSomething()

    static Foo? get sharedInstance => _factorySharedInstance();

    static Function(A,B,C) _factoryMake = Foo$Impl.make
    static Function() _factoryDoSomething = Foo$Impl.doSomething
    static Function() _factorySharedInstance = Foo$Impl.sharedInstance
}

class Foo$Impl extends __lib.NativeBase implements Foo {
    Foo$Impl.make(A a, B b, C c) {
    }

    static SDKNativeEngine? get sharedInstance {
    }

    static void doSomething() {
    }
...
}

```

## Prototype class

Approach with prototype class looks easier and most convenient.

[...]

PR #1642, opened 2024-12-19
Dart: do not derive from classes annotated as 'visibleForTesting'

pwrobeldev:
----- Motivation -----
In order to allow users to mock static functions in classes
the 'prototype' approach was implemented in the past. It uses
prototype class to redirect static function calls.
    
A class with '$Impl' suffix is created as a prototype. It is
annotated as 'visibleForTesting' and exported from the library.
    
The user may provide its own implementation in tests and this
way redirect the calls to the mock (after setting it as prototype).
    
Sadly, when a class is derived from the class, which has static
methods, then its '$Impl' derives from the '$Impl' of base class.
This leads to linter warnings: 'invalid_use_of_visible_for_testing_member'.

----- Implemented solution -----    
This change introduces a new type called '$HiddenImpl', which is
defined only when the class is open and has static functions.
The '$HiddenImpl' is not exported and therefore, the user cannot use it
directly. Still, the user may use the old approach to redirect the calls
to mock.

The '$HiddenImpl' is derived from an ordinary '$Impl' and its usage does
not generate linter warnings when the '$Impl' of derived class uses it,
because '$HiddenImpl' is not annotated as 'visibleForTesting'.
    
The '$HiddenImpl' is used only to provide implementation of methods from
the base class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Gluecodium project Release Notes
 
+## Unreleased
+### Bug fixes:
+ * Dart: removed inheritance from base types annotated as `visibleForTesting`. It caused linter warnings.
+
 ## 13.10.1
 Release date 2024-12-12
 ### Bug fixes:
--- a/functional-tests/functional/CMakeLists.txt
+++ b/functional-tests/functional/CMakeLists.txt
@@ -272,8 +272,10 @@ feature(Defaults cpp android swift dart SOURCES
 )
 
 feature(Inheritance cpp android swift dart SOURCES
+    input/src/cpp/BaseClassWithStaticMethodsImpl.cpp
     input/src/cpp/ChildClassImpl.cpp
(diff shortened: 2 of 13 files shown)


pwrobeldev:
A note for reviewers:
- more information about the warning can be found here: https://dart.dev/tools/diagnostic-messages#invalid_use_of_visible_for_testing_member

The initial design of mock-ability of static functions in Dart can be found here:
- https://github.com/heremaps/gluecodium/issues/1028

pwrobeldev:
After the offline discussion with @Hsilgos I am closing this PR.

The fix looks like a hack. It will be replaced via `ignore` for the given warning to avoid overcomplicating the code.

pwrobeldev:
Just for a reference -- the PR was closed, because the problem is fixed via: https://github.com/heremaps/gluecodium/pull/1644

PR #1650, opened 2025-02-11
Dart: fix compilation error caused by missing includes for async functions

pwrobeldev:
FfiCppIncludeResolver class was not resolving includes correctly,
because certain includes were missing. Sadly, the problem was not
visible because of the order of includes in unity file.

This change:
- adjusted functional tests to reproduce the compilation error caused by missing includes
- implemented fix in FfiCppIncludeResolver
- Adjusted smoke tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Bug fixes:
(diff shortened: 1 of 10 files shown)


pwrobeldev:
Note to reviewers: the functional tests failed for the reproducer (visible in history).

PR #1651, opened 2025-02-11
LimeParser: non-greedy parsing of new lines for ceratin constants

pwrobeldev:
Certain sub-rules of 'constant' rule used greedy
matching of new lines at the end:
 - 'structInitializer'
 - 'listInitializer'
 - 'mapInitializer'

If the input LIME file contained invalid syntax
after such constant definition, then the parser
tried to apply greedy matching. It resulted in
meaningless error message e.g.:
 - `mismatched input '//' expecting NewLine`

This change adds the usage of 'Nongreedy Parser Subrules'
which prevent the greedy matching and meaningless error messages.

### DIFF
--- a/lime-loader/src/main/antlr/LimeParser.g4
+++ b/lime-loader/src/main/antlr/LimeParser.g4
@@ -242,15 +242,15 @@ positionalEnumeratorRef
 
 structInitializer
     : '{' NewLine* ((simpleId NewLine* '=' NewLine*)? literalConstant NewLine*
-      (',' NewLine* (simpleId NewLine* '=' NewLine*)? literalConstant NewLine*)*)? '}' NewLine*
+      (',' NewLine* (simpleId NewLine* '=' NewLine*)? literalConstant NewLine*)*)? '}' NewLine*?
     ;
 
 listInitializer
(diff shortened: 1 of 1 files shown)


pwrobeldev:
From the docs:
 - https://github.com/antlr/antlr4/blob/master/doc/wildcard.md#nongreedy-parser-subrules

> [...] parsers always make globally correct decisions. A parser never makes a decision that will ultimately cause valid input to fail later on during the parse. Here is the central idea: Nongreedy parser subrules match the shortest sequence of tokens that preserves a successful parse for a valid input sentence.

pwrobeldev:
Some minimum reproducer:

```
class SomeClass {
    const LABELS: List<String> = ["abc", "def"]

    // Some valid function.
    fun validFun(): Int

    // Some invalid function.
    fun invalidFun(
        p1: Int,
        p2: Int,
    ): Int
```


The error without the fix points to the `validFunction` without any meaningful context:
```
mismatched input '//' expecting NewLine
```


The output with fix points to `invalidFunction` with context:
```
extraneous input ')' expecting {NewLine, '@', Identifier, '//', '/*'}
```

Hsilgos (approved):

Hsilgos (approved):

PR #1665, opened 2025-02-25
Kotlin: support usage of external types

pwrobeldev:
This commit introduces a set of test cases for
usage of external types for Kotlin generated code.
It also fills the gaps in the existing implementation
to make the tests pass.
    
Two important notes:
 - fundamental numeric types of Kotlin are deliberately
 used with '?' suffix when defining the external types,
 because currently JNI templates assume that the used type
 is an object
 - for fundamental types this assumption is true only if
  the nullable version is used; for instance 'Kotlin.Int'
  is mapped to trivial 'int' whether 'Kotlin.Int?' is mapped
  to 'java.lang.Integer'
    
The support for usage of non-nullable optional types as the
external types will be added in a follow-up PR, because it
requires changes in JNI mustache files

### DIFF
--- a/functional-tests/functional/CMakeLists.txt
+++ b/functional-tests/functional/CMakeLists.txt
@@ -303,7 +303,7 @@ feature(CircularDependencies cpp android swift dart SOURCES
     input/lime/Circular.lime
 )
 
-feature(ExternalTypes cpp android swift dart SOURCES
+feature(ExternalTypes cpp android android-kotlin swift dart SOURCES
     input/src/cpp/include/ExternalTypes.h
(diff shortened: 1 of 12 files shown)


pwrobeldev:
A few notes for reviewers:
- before this change can be reviewed the following one must be merged and this branch needs to be rebased and adjusted: https://github.com/heremaps/gluecodium/pull/1664
- this PR does not introduce support for usage of non-nullable fundamental Kotlin types -- this will be done as a follow-up work, because it requires changes in JNI layer

Hsilgos (approved):

PR #1666, opened 2025-02-26
Implement placeholders for comments

pwrobeldev:
----- Motivation -----
When creating documentation comments there are elements, which may be
used often. Sometimes the repetition leads to problems. Let's look at the following
example showing the code needed to properly word the `interface` for different outputs:

```
{@Java interface}{@Swift protocol}{@Dart abstract class}{@Cpp abstract class}
```

It is easy to forget adding wording for one of output platforms. Moreover, the mistake
may be repeated if this long syntax is copied to other places in the code.

----- Solution -----
The placeholders for comments are introduced to solve the described problem.
Instead of using the long syntax the new `{@Placeholder <PLACEHOLDER_NAME>}`
can be used.

The new configuration parameters, which allow setting placeholders file are available:
 - CLI: `-docsplaceholderslist`
 - CMake: `GLUECODIUM_DOCS_PLACEHOLDERS_LIST`

The file must follow `.properties` file format and must contain the known comment placeholders
as `placeholder_name="placeholder_value"`.

Note: the placeholders respect the usual LIME markup -- it means that nested platforms tag can
be used, but they are not required.

### DIFF
--- a/cmake/tests/utils/run_check_file_contains_string.cmake
+++ b/cmake/tests/utils/run_check_file_contains_string.cmake
@@ -19,7 +19,7 @@ cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 include("${CMAKE_CURRENT_LIST_DIR}/message_colored.cmake")
 
-set(_required_vars CHECK_FILE_DOES_NOT_CONTAIN_FILE_PATH CHECK_FILE_DOES_NOT_CONTAIN_NEEDLE)
+set(_required_vars CHECK_FILE_CONTAINS_FILE_PATH CHECK_FILE_CONTAINS_NEEDLE)
 
 foreach(_required_variable ${_required_vars})
   if(NOT DEFINED ${_required_variable})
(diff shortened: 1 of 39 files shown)


cmake/tests/utils/run_check_file_contains_string.cmake
```diff
@@ -0,0 +1,47 @@
...
+
+if(NOT EXISTS ${CHECK_FILE_CONTAINS_FILE_PATH})
+    message_colored(RED FATAL_ERROR "${CHECK_FILE_CONTAINS_FILE_PATH} does not exist!")
+endif()
+
+file(STRINGS ${CHECK_FILE_CONTAINS_FILE_PATH} _lines)
+
+set(FOUND_NEEDLE FALSE)
+
+foreach(_line ${_lines})
```
Hsilgos, line 38: `foreach (_line IN LISTS _lines)` is safer a bit

pwrobeldev: Sure, it has been fixed and force-pushed 👍

Hsilgos (approved):

PR #1716, opened 2025-04-23
Conditionally warn about android attributes mismatch

pwrobeldev:
In order to ease transition from Java to Kotlin
the users have the possibility to enable warnings
about mismatch in attributes for elements.
    
The following new CLI parameter is available
`-enableandroidattributesmismatchwarning` as well as
`GLUECODIUM_ENABLE_ANDROID_ATTRIBUTES_MISMATCH_WARNING`
CMake parameter.
    
When flag is set to true and an element from LIME file uses
annotations for Java, but it does not use the matching annotation
for Kotlin then the warning is generated (and vice versa).
    
Signed-off-by: Patryk Wrobel <183546751+pwrobeldev@users.noreply.github.com>

### DIFF
--- a/cmake/tests/unit/gluecodium_generate/enable-android-attributes-mismatch-warning-XFAIL/successful.regex
+++ b/cmake/tests/unit/gluecodium_generate/enable-android-attributes-mismatch-warning-XFAIL/successful.regex
@@ -1,0 +1,1 @@
+element unit.test.Foo: Attributes missing in Kotlin, but present in Java: \[Skip\]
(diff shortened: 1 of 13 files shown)


cmake/tests/unit/gluecodium_generate/enable-android-attributes-mismatch-warning-XFAIL/successful.regex
```diff
@@ -0,0 +1 @@
+element unit.test.Foo: Attributes missing in Kotlin, but present in Java: \[Skip\]
```
Hsilgos, line 1: I don't get why script fails if warning is not treated as error. Did I miss something?

pwrobeldev: I had to force failure of the script by passing invalid LIME file to see the logs that are produced. Please see the comment in foo.lime:

```
# This class uses function with type that does not exist.
# It is done this way to force CMake to print Gluecodium logs
# to the output to be able to match it.
@Java(Skip)
class Foo {
    fun foo(): SomeTypeThatDoesNotExits
}
```

Hsilgos (approved):

PR #1807, opened 2026-06-03
CI: ensure that macOS runner is x64

pwrobeldev:
---------- Motivation ----------
The runner images for 'macOS 13' had the following convention regarding the hardware architecture:
 - 'macos-13' --> x64
 - 'macos-13-xlarge' --> arm64

It has been different since 'macOS 14':
 - 'macos-14' --> arm64
 - 'macos-14-large' --> x64
 - 'macos-15' --> arm64
 - 'macos-15-intel' and 'macos-15-large' --> x64

In January 2026 the runner image was updated from 'macos-13' to 'macos-14' because the older
version was deprecated. It was not checked that the runner architecture is different -- the change
in runner naming convention was not expected.

Recently the job, which executes functional tests for Swift generated code on macOS started to fail.
The installation of Java did not work properly. The investigation revealed that Java was installed for
x64 instead of arm64. The action used to install Java cannot be configured for arm.

---------- Solution ----------
Introduced the usage of 'macos-15-intel' runner to ensure that the hardware architecture of the runner is x64.
Moreover, adjusted the environment variable to avoid problems with 'JAVA_HOME' variable.

### DIFF
--- a/.github/workflows/cmake-tests-macos.yml
+++ b/.github/workflows/cmake-tests-macos.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   cmake-tests-ios:
     name: CMake toolchain tests for iOS
-    runs-on: macos-14
+    runs-on: macos-15-intel
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


pwrobeldev:
The CMake tests on macOS were also executed with the updated CI configuration. They passed:
 - https://github.com/heremaps/gluecodium/actions/runs/26867002298

Hsilgos (approved):

PR #1808, opened 2026-06-23
WIP: Kotlin Coroutines

pwrobeldev:
POC


functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
```diff
@@ -0,0 +1,148 @@
...
+import org.robolectric.annotation.Config
+
+import kotlinx.coroutines.*
+
+data class SuspendingJobResult(val labels: List<String>, val timestamp: Long, val exceptionCaught: Boolean)
+
+@RunWith(RobolectricTestRunner::class)
+@Config(application = RobolectricApplication::class)
+class AsyncDecoratorTest {
+    data class SuspendingJobResult(val labels: List<String>, val timestamp: Long)
```
HERE-SDK-Support-Team: redeclaration?

functional-tests/functional/input/src/cpp/AsyncDecorator.cpp
```diff
@@ -0,0 +1,184 @@
...
+        if (do_work)
+        {
+            m_worker = std::thread{
+                [this, w = std::move(do_work)] {
+                    w();
+                    m_is_running.store(false);
+                }
+            };
+
+            m_is_running.store(true);
```
HERE-SDK-Support-Team: m_is_running is set to true after launching the worker thread. If the worker finishes quickly, it can set false first, then constructor sets true, leaving a stale running state?

functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
```diff
@@ -0,0 +1,148 @@
...
+            assertTrue(t2Result.labels.isEmpty())
+            assertTrue(t2Result.exceptionCaught)
+            // Check that job took at least 500ms, and less than 1000ms (margin).
+            val t2ElapsedTimeMs = t2Result.timestamp - startTimestamp
+            assertTrue("Time of the second job: $t2ElapsedTimeMs >= 500ms", t2ElapsedTimeMs >= 500)
+            assertTrue("Time of the second job: $t2ElapsedTimeMs < 1000ms", t2ElapsedTimeMs < 1000)
+        }
+
+        // Ensure that underlying C++ threads join.
+        engine.waitForCompletion()
```
HERE-SDK-Support-Team: If any assertion fails earlier, cleanup is skipped?

functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
```diff
@@ -0,0 +1,148 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+package com.here.android.test
```
HERE-SDK-Support-Team, line 19: - AsyncDecorator(CancelFunction="cancel") not tested?
- not all EngineError tested?

functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
```diff
@@ -0,0 +1,148 @@
...
+        // Important note:
+        //  - please be aware that final expectations calculate elapsed time between the beginning of the test case and
+        //    completion of given coroutine
+        //  - we assume that concurrent tasks start immediatelly after the start of the test case
+        //  - this way we can tell that they were executed concurrently and there was no blocking -- if there was blocking
+        //    then elapsed time would be total of given task and the ones before it
+        runBlocking {
+            // This task should work for a bit longer than 2000 ms.
+            val t0 = async {
+                // Worker thread on C++ level will sleep for: 1000 ms.
```
Hsilgos, line 76: 1 and 3 seconds is too much for unit tests. 100ms - 300ms is better, but still timeouts should be more accurate

functional-tests/functional/input/lime/AsyncDecorator.lime
```diff
@@ -0,0 +1,64 @@
...
+
+// An engine class, which offers users some asynchronous processing.
+class CoolEngine {
+    // Creates engine ready to work.
+    constructor make()
+
+    // A function, which downloads some 'cool labels' from the given url.
+    // When it finishes, the 'callback' is called.
+    // It returns the 'AsyncTaskHandle' which allows checking the status and cancelling the async task.
+    @AsyncDecorator(CancelFunction="cancel")
```
Hsilgos: If you say that it should be specified everywhere may be then make it to specify once on `AsyncTaskHandle`?

functional-tests/functional/input/src/cpp/AsyncDecorator.cpp
```diff
@@ -0,0 +1,184 @@
...
+    std::atomic_bool m_is_running{false};
+};
+
+class CoolEngineImpl : public CoolEngine
+{
+public:
+    std::shared_ptr<test::AsyncTaskHandle>
+    download_cool_labels_async(const ::std::string& url, const ::test::EngineWorkCompletedCallback& callback) override
+    {
+        std::function<void()> do_work = [url, c = std::move(callback)] {
```
Hsilgos: `callback = std::move(callback)` works well

pwrobeldev: I think that I made mistake -- callback is const reference. Thus, the move does not make much sense.

functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
pwrobeldev: Yes, my bad -- I was going fast when creating the POC. 

It will be removed 👍

functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
pwrobeldev: The cleanup is also present in the destructor of C++ object. So, technically it is always called. But I see the possible problem if the object was destroyed from another thread -- that's why I wanted to add it here.

Thanks for pointing this out -- the code will be adjusted 👍

functional-tests/functional/input/src/cpp/AsyncDecorator.cpp
pwrobeldev: Yeah, this was wrong. Fixed.

functional-tests/functional/input/src/cpp/AsyncDecorator.cpp
pwrobeldev: Fixed -- the callback is captured via full name (`callback`) and by value -- move did not help, because of const ref.

functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
pwrobeldev: Done --> reworked to use setUp/tearDown via `@Before` and `@After`.

functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
pwrobeldev: Fixed according to the comment.

functional-tests/functional/input/lime/AsyncDecorator.lime
pwrobeldev: Done --> `@AsyncTaskHandleMetadata` annotation was added

functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
pwrobeldev: I restored the original delays.

When I decreased the times 10x, then the margin of 50ms was too small for one of assertions (visible only on CI).

On my local machine additional execution time was ~30ms. On CI it was more than 50ms.

Thus, I restored the initial delays -- still, the logic is tested, because if the tasks were not executed concurrently and thread was blocked, then timeout of t2 would be t1+t2, and t3 would be t1+t2+t3

functional-tests/functional/android-kotlin/src/test/kotlin/com/here/android/test/AsyncDecoratorTest.kt
pwrobeldev: The additional test case for cancellation has been added.

pwrobeldev:
I am closing the PR -- this was a POC. The proper PR will be created with the final solution.

PR #1809, opened 2026-06-29
fix(dart): hide internal fields in ctor

isidorowiczdev:
Fix Dart generator so internal fields stay out of public all-args constructor. Reason: current generator leaks @Internal field into public API, unlike Java, and smoke tests show broken output.


pwrobeldev:
As discussed offline -- besides the problem with `all args ctor` there seems to also be a problem when we have at least 1 initialized field in the structure.

Please have a look at the following input from our smoke tests:
 - https://github.com/heremaps/gluecodium/blob/master/gluecodium/src/test/resources/smoke/visibility_attribute/input/VisibilityInternal.lime#L94

And the corresponding output for Dart:
 - https://github.com/heremaps/gluecodium/blob/master/gluecodium/src/test/resources/smoke/visibility_attribute/output/dart/lib/src/smoke/public_struct_with_non_default_internal_field.dart#L9

In the output file you can see that `all args ctor` is private, but `initialized fields ctor` that is public exposes the internal field and takes its value -- this does not seem correct.

Hsilgos:
Hide whole constructor because of one field?
May be filter it out like fields with default values?

gluecodium/src/main/java/com/here/gluecodium/generator/dart/DartNameResolver.kt
```diff
@@ -138,7 +138,7 @@ internal class DartNameResolver(
                     when {
                         !useDefaultsConstructor -> ""
                         noFieldsConstructor == null ->
-                            if (DartGeneratorPredicates.allFieldsCtorIsPublic(actualType)) ".withDefaults" else ""
+                            if (DartGeneratorPredicates.allFieldsCtorIsPublic(actualType)) ".withDefaults" else "._withDefaults"
```
pwrobeldev: Unfortunately, this breaks the functional tests. They do not compile. We need to analyze the logic of generation constructors, because the fix may not be trivial.

pwrobeldev:
> Hide whole constructor because of one field? May be filter it out like fields with default values?

@Hsilgos -- is the proposal to filter the internal field from the constructor's parameters list?

If that's the proposal, then I have one question:
- How will that field be initialized if it does not have default value set?

If I understand correctly, if we filtered such field from the constructor then it would be uninitialized.

gluecodium/src/test/resources/smoke/visibility_attribute/input/ConstructorVisibilityWithInternalFields.lime
```diff
@@ -0,0 +1,108 @@
...
+//
+@Java(PositionalDefaults)
+@Kotlin(PositionalDefaults)
+@Dart(PositionalDefaults)
+struct PosDefaultsWithOneInternalField {
+    @Internal
+    internalField: String
+    publicField: Boolean
+    initializedField: Int = 77
+}
```
pwrobeldev: This case will be fixed by a separate PR. Please remove it -- it showcases invalid behavior and this PR does not provide fix for it.

Please also remove the corresponding output files for smoke tests (related to this type).

gluecodium/src/test/resources/smoke/visibility_attribute/input/ConstructorVisibilityWithInternalFields.lime
pwrobeldev, line 1: Please add also the same test cases to functional tests to confirm that the code compiles.

If possible, please add also the test which creates such types and performs round trip (Dart -> C++ -> Dart) to ensure that we do not break the functionality.

pwrobeldev (commented):
In general the change looks good to me.

Please adjust the smoke tests and add the corresponding examples in functional tests.

pwrobeldev:
Note: the change-log is missing.

pwrobeldev:
The change looks good to me.

Hsilgos (approved):

pwrobeldev (approved):

PR #1810, opened 2026-07-08
docs: document all-uppercase acronym transformation

isidorowiczdev:
## Uppercase acronyms in name transformation
Problem: All-uppercase acronyms like `FACTORY` are converted to mixed case (`Factory`) when transformed to camelCase format. This limitation affects users defining single-word uppercase identifiers in LIME.

Solution: Documented the behavior with:
- Root cause: Recognition logic treats all-uppercase single words as generic strings, not recognized patterns
- Workaround: Platform-specific `Name` attributes override transformation rules completely
  - `@Java(Name = "FACTORY")`, `@Kotlin(Name = "FACTORY")`, `@Swift(Name = "FACTORY")`, `@Dart("FACTORY")`

### DIFF
--- a/docs/naming_conventions.md
+++ b/docs/naming_conventions.md
@@ -233,3 +233,52 @@ error.suffix=Exception
 join.infix=_
 
 ```
+
+Handling all-uppercase acronyms
+-------------------------------
+
+### Known limitation
+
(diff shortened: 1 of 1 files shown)


pwrobeldev (approved):
LGTM

PR #1812, opened 2026-08-19
 Add Kotlin coroutine suspend wrappers for @AsyncDecorator functions

tusharpandey1993:
## Summary

Adds Kotlin coroutine support to Gluecodium's code generator, enabling `@AsyncDecorator`-annotated LIME methods to produce idiomatic `suspend` extension functions that bridge callback-based async APIs to Kotlin coroutines.

### What's included

- **New `@AsyncDecorator` LIME attribute**  marks async methods for coroutine wrapper generation
- **Kotlin code generator extensions** produces `suspend` extension functions returning typed results, bridged via a generic `awaitResult` helper
- **Validator** (`LimeAsyncDecoratorValidator`) validates decorator signatures at generation time, rejects duplicates and unsupported patterns
- **Name resolver support**  unique per-package helper names, custom name overrides via attribute
- **Mustache templates**  `KotlinCoroutines.mustache` and `KotlinCoroutineMembers.mustache` for structured output
- **Functional tests**  Android/Kotlin tests with C++ backing implementations covering success, error, and cancellation paths
- **Documentation**  `docs/async.md` with usage guide and attribute reference

### Design decisions

- Scoped to `suspend` wrappers only (Flow support was explored and intentionally dropped to limit scope)
- Extension functions avoid polluting generated class APIs
- Result return types clarify success/failure semantics for callers


pwrobeldev (approved):
This change was reviewed internally and had a few iterations with fixes. I am approving the change.

Please prepare follow-up PRs with:
 1. Adjustment of `CHANGELOG.md` file to inform about the new feature.
 2. More functional tests (e.g. with nested class that has asynchronous functions decorated via `@AsyncDecorator`).

Moreover, we need to handle the case when `@Internal` function is annotated with `@AsyncDecorator`. In such case we should not generate the wrapper as `public` functions but as `internal`.

We could also initially disallow annotating `@Internal` functions via `@AsyncDecorator` and work on this later.

PR #1814, opened 2026-08-21
Kotlin: improve @AsyncDecorator callback selection syntax and coroutine file naming

tusharpandey1993:
## Summary

- Callback selection is now declared on the function (`@AsyncDecorator(Callback = "paramName")`)
  instead of on the parameter (`@AsyncDecorator(Callback)`). This is cleaner and removes ambiguity
  when a function has multiple lambda parameters.
- Generated KotlinCoroutines files now use readable names like `KotlinCoroutines_Route` instead of
  opaque hashes like `KotlinCoroutines_8e0c10ac`.
- Added functional tests for `@AsyncDecorator` on nested classes.

## Test plan

- [x] All gluecodium unit/smoke tests pass
- [x] Functional tests pass (nested class success + error paths)


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Unreleased
  * C++: introduced the new 'EnumValues' attribute, which allows uses to instruct the C++ generator to create helper function for the given enumeration. The mentioned function returns an array of unique enumerator values.
  * Kotlin: added `@AsyncDecorator` attribute for generating Kotlin coroutine `suspend` extension functions from callback-based asynchronous functions. This allows Kotlin consumers to call async APIs using idiomatic coroutines instead of raw callbacks. Supports result/error mapping, cancellation via `@AsyncTaskHandle`, and custom wrapper naming via `@AsyncDecorator(Name = "...")`.
+ * Kotlin: moved `@AsyncDecorator` callback selection from parameter-level to function-level syntax (`@AsyncDecorator(Callback = "paramName")`), enabling clearer intent when a function has multiple lambda parameters.
+ * Kotlin: replaced opaque hex-hash suffixes in generated `KotlinCoroutines` JVM file-class names with human-readable container names (e.g. `KotlinCoroutines_Route` instead of `KotlinCoroutines_8e0c10ac`).
```
pwrobeldev: I would merge the messages related to Kotlin coroutines into one clear message for the users. 

This note will be present in release notes of the given Gluecodium version. 

For the user only the final state is important -- we do not need to describe the steps that override each other.

gluecodium/src/main/java/com/here/gluecodium/generator/kotlin/KotlinAsyncHelpers.kt
```diff
@@ -65,9 +65,8 @@ internal object KotlinAsyncHelpers {
      * file-class name this produces duplicate `.../KotlinCoroutinesKt` classes during Kotlin compilation.
      */
     private fun coroutineFileJvmSuffix(containers: List<LimeContainer>): String {
-        val signature = containers.map { it.path.toString() }.sorted().joinToString("|")
-        val hex = signature.hashCode().toUInt().toString(16)
-        return "_$hex"
+        val names = containers.map { it.path.container }.distinct().sorted()
+        return "_${names.joinToString("_")}"
```
pwrobeldev: 1. Is there some limit for the maximum length of JVM class name? The annotation: `JvmFile()` is used to create the class name that encapsulates the free standing functions defined in the file.
2. Isn't the generated class name a bit strange?

Lastly, can the current hash-based and the proposed class-name based approach create problems with binary compatibility? Does it mean that adding a new class with async decorator causes breaking ABI change because the class that existed in the past will have a new name?

CHANGELOG.md
tusharpandey1993: Done

gluecodium/src/main/java/com/here/gluecodium/generator/kotlin/KotlinAsyncHelpers.kt
tusharpandey1993: Now emitting one coroutine file per top-level container (KotlinCoroutines_Route.kt), with nested classes grouped into their parent's file. Each file carries a stable @file:JvmName("KotlinCoroutines_<ContainerName>"), so adding/removing containers in the same package never renames an existing JVM class. No ABI breakage.

pwrobeldev (approved):

PR #1816, opened 2026-08-26
C++14 support: restore usage of type alias for optional type

pwrobeldev:
---------- Problem context ----------
The generated C++ code and bindings must be compilable with C++14.
Multiple places in the code used `std::optional` type directly. This type
is available in C++17 or newer.

---------- Implemented fix ----------
1. Reintroduced the custom optional type generation to Gluecodium.
2. Reintroduced optional type alias which conditionally resolves to either custom type or std::optional.
3. Adjusted functional, smoke, and cmake tests to utilize the type alias.
4. Adjusted the example application to use the type alias.
5. Adjusted the documentation files with developer guide.

Note -- this change is partial revert of the following PRs:
 - https://github.com/heremaps/gluecodium/pull/1427
 - https://github.com/heremaps/gluecodium/pull/1466


pwrobeldev:
With this PR the only files, which mention `std::optional` are the following:

```
./CHANGELOG.md
./docs/guide.md
./gluecodium/src/main/resources/templates/cpp/common/Optional.mustache
./gluecodium/src/main/resources/templates/cpp/common/Return.mustache
```

The first two files are documentation related.

The `Optional.mustache` is template for generating type alias and `std::optional` is used under `ifdef`.

The `Return.mustache` mentions `std::optional` in documentation comment to give the users idea about the type -- it does not use the implementation.

pwrobeldev:
Note for reviewers: the changes are done commit-by-commit to ease the review.

Hsilgos (dismissed):

Hsilgos (approved):

PR #1817, opened 2026-09-01
C++14 support: generate std::string_view replacement for older standards

pwrobeldev:
---------- Problem context ----------
The generated C++ code and bindings must be compilable with C++14.
Multiple places in the code used std::string_view type directly. This type
is available in C++17 or newer.

---------- Implemented fix ----------
1. Introduced the custom 'StringView' type generation to Gluecodium.
2. Introduced string view type alias which conditionally resolves to either custom type or std::string_view.
3. Adjusted smoke tests to utilize the type alias.
4. Added the changelog note.


pwrobeldev:
With the changes from the PR the only places that explicitly mention `std::string_view` are:

```
./CHANGELOG.md
./gluecodium/build/resources/main/templates/cpp/common/StringViewImpl.mustache
./gluecodium/build/resources/main/templates/cpp/common/StringView.mustache
```

gluecodium/src/main/resources/templates/cpp/common/StringViewImpl.mustache
```diff
@@ -0,0 +1,731 @@
...
+    constexpr void
+    remove_suffix(size_type n)
+    {
+        m_size = (m_size - n);
+    }
+
+    constexpr void
+    swap(StringView& other) noexcept
+    {
+        // Do not call std::swap(). It is not constexpr in C++14.
```
Hsilgos, line 260: But you _may_ implement constexpr swap ;)

Hsilgos (approved):

PR #1820, opened 2026-09-02
Kotlin: generate Kotlin data classes for tagged structs

tusharpandey1993:
## Description

Adds a `@Kotlin(DataClass)` attribute that makes Gluecodium emit a LimeIDL `struct` as an immutable Kotlin `data class` (`val` members, public constructor, `copy()`/`equals()`/`hashCode()`/`componentN()`), instead of the existing NativeBase-wrapper style. Struct methods keep being generated as external functions and statics in a companion object.

The attribute is intentionally restricted to `struct`; applying it to anything else is rejected at build time by a dedicated validator, since a non-`struct` element's Kotlin/JNI construction path (`create_instance_object`, expecting a `(Long, Any?)` constructor) is incompatible with the data-class constructor shape and would otherwise crash at runtime.

`@Kotlin(DataClass)` also requires the struct to be marked `@Immutable`, and requires that it does not declare any explicit `constructor`/`field constructor` (since the data class' primary constructor is always generated from the struct's fields, so a hand-written constructor would otherwise be silently ignored). Both of these are enforced by the same validator.

## Changes
- `lime-runtime`/`lime-loader`: register and parse the `DataClass` attribute value.
- `KotlinGenerator`: select the data-class template path for tagged structs.
- `KotlinStruct.mustache`: emit `data class` with a `val` primary constructor, including per-field doc comments, attributes, and resolved visibility.
- `LimeKotlinDataClassValidator` (+ tests): checks every named element (not just `LimeClass`) for the `DataClass` attribute; hard build error if it's applied to anything other than a `struct`, if the struct isn't also `@Immutable`, or if the struct declares an explicit constructor or field constructor.
- `LimeAndroidAttributesMismatchValidator` / `GeneratorOptions` / `OptionReader`: replaced the hardcoded `DataClass` exclusion from the Java/Kotlin attribute-mismatch check with a configurable `androidAttributesMismatchAllowlist` option (new `-androidattributesmismatchallowlist` CLI flag), defaulting to `[DataClass]` to preserve current behavior.
- New `data_class` smoke-test suite with `Place`/`Suggestion` fixtures, including LIME doc comments to verify KDoc generation end-to-end.
- `CHANGELOG.md` entry under Unreleased.

## Testing
- `./gradlew spotlessApply`
- `./gradlew :lime-runtime:build :lime-loader:build :gluecodium:build :gluecodium-gradle:build` — BUILD SUCCESSFUL


gluecodium/src/main/java/com/here/gluecodium/generator/kotlin/KotlinGenerator.kt
```diff
@@ -295,7 +296,12 @@ internal class KotlinGenerator : Generator {
 
     private fun selectTemplate(limeElement: LimeNamedElement) =
         when (limeElement) {
-            is LimeClass -> "kotlin/KotlinClass"
+            is LimeClass ->
+                if (limeElement.attributes.have(KOTLIN, DATA_CLASS)) {
+                    "kotlin/KotlinDataClass"
+                } else {
+                    "kotlin/KotlinClass"
+                }
```
pwrobeldev, line 304: It looks like some leftover from the initial development and is not aligned with description and smoke tests.

The description and smoke tests mention generation of `LimeIDL struct` -- not class.

> Adds Kotlin(DataClass) attribute that makes Gluecodium emit a LimeIDL struct as an immutable Kotlin data class (val members, public constructor, copy()/equals()/hashCode()/componentN()), instead of the existing NativeBase-wrapper style. Struct methods keep being generated as external functions and statics in a companion object.

gluecodium/src/main/java/com/here/gluecodium/validator/LimeAndroidAttributesMismatchValidator.kt
```diff
@@ -47,7 +47,8 @@ class LimeAndroidAttributesMismatchValidator(private val limeLogger: LimeLogger,
         val commonAttributes = kotlinAttributes intersect javaAttributes
 
         var result = true
-        val attributesMissingInJava = kotlinAttributes subtract commonAttributes
+        val attributesMissingInJava =
+            (kotlinAttributes subtract commonAttributes) - LimeAttributeValueType.DATA_CLASS
```
pwrobeldev, line 51: Instead of hardcoding it here we should allow the users to set the list of parameter types that can be misaligned. 

The validation of consistency between Java and Kotlin was introduced to ease the migration from Java to Kotlin.

gluecodium/src/main/java/com/here/gluecodium/validator/LimeKotlinDataClassValidator.kt
```diff
@@ -0,0 +1,66 @@
...
+ * generated C++ still calls `create_instance_object`, which fails to find the `(long, Object)`
+ * constructor at runtime and crashes with `NoSuchMethodError` inside `JNI NewObjectV`. This is a
+ * deterministic runtime crash, not a build-time error, so it must be caught here instead.
+ */
+internal class LimeKotlinDataClassValidator(private val logger: LimeLogger) {
+    fun validate(limeModel: LimeModel): Boolean {
+        val offendingClasses =
+            limeModel.referenceMap.values
+                .filterIsInstance<LimeClass>()
+                .filter { it.attributes.have(KOTLIN, DATA_CLASS) }
```
pwrobeldev, line 50: Instead of filtering only `LimeClass` instances I would check all elements and if we see that `DATA_CLASS` attribute is specified, then we should check if the element is LimeStruct.

gluecodium/src/main/resources/templates/kotlin/KotlinDataClass.mustache
pwrobeldev, line 1: If I understand correctly this is a leftover and should be removed.

gluecodium/src/main/resources/templates/kotlin/KotlinStruct.mustache
```diff
@@ -24,7 +24,15 @@
 {{/unless}}{{/internalApiAnnotation}}internal {{!!
 }}{{/if}}{{!!
 }}{{#unless external.kotlin.converter}}{{>kotlin/KotlinTypeVisibility}}{{/unless}}{{!!
-}}class {{resolveName}}{{!!
+}}{{#if attributes.kotlin.dataClass}}data {{/if}}class {{resolveName}}{{!!
+}}{{#if attributes.kotlin.dataClass}}(
+{{#fields}}    val {{resolveName}}: {{resolveName typeRef}}{{#if defaultValue}} = {{resolveName defaultValue}}{{/if}}{{#if iter.hasNext}},{{/if}}
```
pwrobeldev, line 29: Fields must be documented and must use proper visibility.

gluecodium/src/main/java/com/here/gluecodium/validator/LimeKotlinDataClassValidator.kt
```diff
@@ -0,0 +1,66 @@
...
+ * constructed via `create_instance_object`, which looks up the NativeBase constructor with signature
+ * `(long, Object)`.
+ *
+ * A Kotlin `data class` generated from `KotlinDataClass.mustache` has only the fields-based primary
+ * constructor and no `(Long, Any?)` constructor. If `@Kotlin(DataClass)` is applied to a `class`, the
+ * generated C++ still calls `create_instance_object`, which fails to find the `(long, Object)`
+ * constructor at runtime and crashes with `NoSuchMethodError` inside `JNI NewObjectV`. This is a
+ * deterministic runtime crash, not a build-time error, so it must be caught here instead.
+ */
+internal class LimeKotlinDataClassValidator(private val logger: LimeLogger) {
```
pwrobeldev, line 45: In the current behavior `DataClass` overrides other constructors.

Should we check if there are no constructors defined?

pwrobeldev: Moreover, the assumption is that we create immutable class -- do we need to require both attributes?

gluecodium/src/test/resources/smoke/data_class/input/DataClassStructs.lime
```diff
@@ -0,0 +1,59 @@
...
+    id: String
+    place_type: PlaceType
+    latitude: Double
+    longitude: Double
+    distance_in_meters: Int? = null
+    political_view: String? = null
+
+    fun serialize_compact(): String
+
+    static fun deserialize(serialized_place: String): Place
```
pwrobeldev, line 40: Please add documentation comments to LIME to check if proper documentation comments are generated for the output.

gluecodium/src/test/resources/smoke/data_class/input/DataClassStructs.lime
```diff
@@ -0,0 +1,59 @@
...
+
+enum SuggestionType {
+    CATEGORY,
+    CHAIN,
+    PLACE
+}
+
+@Immutable
+@Kotlin(DataClass)
+struct Suggestion {
```
pwrobeldev, line 51: If this structure defines constructor or field constructor, then with the current state of this branch the generation will be omitted silently (without warnings or errors).

I doubt that it is the desired behavior.

gluecodium/src/main/java/com/here/gluecodium/generator/kotlin/KotlinGenerator.kt
tusharpandey1993: Fixed: removed the leftover `LimeClass` + `DATA_CLASS` branch in `selectTemplate()`; classes now always route to `kotlin/KotlinClass` regardless of the attribute, matching the PR description and smoke tests (struct-only).

gluecodium/src/main/resources/templates/kotlin/KotlinDataClass.mustache
tusharpandey1993: Fixed: removed `KotlinDataClass.mustache`, it was unreachable once the leftover `LimeClass` routing above was removed.

gluecodium/src/main/java/com/here/gluecodium/validator/LimeAndroidAttributesMismatchValidator.kt
tusharpandey1993: Fixed: added a new `androidAttributesMismatchAllowlist` option (in `GeneratorOptions`, wired via new CLI flag `-androidattributesmismatchallowlist`) that users can configure to any set of attribute-value tags they want excluded from the mismatch check. Defaults to `[DataClass]` to preserve current behavior; passing an empty/overridden list re-enables the warning for `DataClass` mismatches too.

gluecodium/src/main/resources/templates/kotlin/KotlinStruct.mustache
tusharpandey1993: Fixed: the data-class primary-constructor field block in `KotlinStruct.mustache` now renders the field's KDoc comment, attributes, and resolved visibility (matching the same partials used for the non-data-class field path) before each `val` declaration.

gluecodium/src/test/resources/smoke/data_class/input/DataClassStructs.lime
tusharpandey1993: Fixed: added LIME doc comments to `Place`, `Suggestion`, their enums, fields, and functions in `DataClassStructs.lime`, and updated the expected smoke-test outputs to include the generated KDoc.

gluecodium/src/main/java/com/here/gluecodium/validator/LimeKotlinDataClassValidator.kt
tusharpandey1993: Fixed: `LimeKotlinDataClassValidator` now iterates over all named elements in the model, filters for those carrying `@Kotlin(DataClass)`, and then checks whether each one `is LimeStruct` (reporting an error otherwise) instead of only ever looking at `LimeClass` instances.

gluecodium/src/main/java/com/here/gluecodium/validator/LimeKotlinDataClassValidator.kt
tusharpandey1993: Fixed: the validator now rejects any `@Kotlin(DataClass)` struct that declares an explicit `constructor` or `field constructor`, since the primary constructor is always generated from the struct's fields and any user-defined constructor would otherwise be silently ignored for Kotlin.

tusharpandey1993: Went with requiring both attributes explicitly: the validator now also requires `@Immutable` to be present alongside `@Kotlin(DataClass)`, since a Kotlin data class only ever has `val` properties and this keeps the Kotlin representation consistent with what C++/Swift/Dart would also need to treat as immutable.

gluecodium/src/test/resources/smoke/data_class/input/DataClassStructs.lime
tusharpandey1993: Fixed: this is now rejected at build time by `LimeKotlinDataClassValidator` (see the constructor-related fix above) instead of being silently dropped. The current `DataClassStructs.lime` in this PR doesn't declare any constructors, so it isn't exercising that failure path itself, but `LimeKotlinDataClassValidatorTest` now covers it directly.

